PR #76, opened 2023-05-11
strip response so binary textcat works

kabirkhan:
## Description

The way our textcat prompt is setup, the LLM (at least OpenAI API) will return a response with `\n\n` at the beginning.

So we get e.g. `\n\nPOS` for binary classification but we aren't stripping the response. So currently we'll always get a 0.0 classification for binary textcat (1 label and exclusive_classes = True)

### Types of change

bug fix

## Checklist

### DIFF
--- a/spacy_llm/tasks/textcat.py
+++ b/spacy_llm/tasks/textcat.py
@@ -137,6 +137,7 @@ Text:
         The returned dictionary contains the labels mapped to their score.
         """
         categories: Dict[str, float]
+        response = response.strip()
         if self._use_binary:
             # Binary classification: We only have one label
             label: str = list(self._label_dict.values())[0]
--- a/spacy_llm/tests/tasks/test_textcat.py
+++ b/spacy_llm/tests/tasks/test_textcat.py
@@ -199,6 +199,8 @@ def test_textcat_sets_exclusive_classes_if_binary():
         ("Some test text with weird response", "WeIrD OUtpuT", 0.0),
         ("Some test text with lowercase response", "pos", 1.0),
         ("Some test text with lowercase response", "neg", 0.0),
+        ("Some test text with unstripped response", "\n\n\nPOS", 1.0),
+        ("Some test text with unstripped response", "\n\n\nNEG", 0.0),
     ],
 )
 def test_textcat_binary_labels_are_correct(text, response, expected_score):


spacy_llm/tasks/textcat.py
```diff
@@ -137,6 +137,7 @@ def _format_response(self, response: str) -> Dict[str, float]:
         The returned dictionary contains the labels mapped to their score.
         """
         categories: Dict[str, float]
+        response = response.strip()
```
ljvmiranda921, line 140: I'm ok with this solution! Thank you!

Optional suggestion: I'm wondering if there's a way we can reuse the normalizer function (that also does the stripping) here? 

Let's also add a test case in "test_binary_labels_are_correct"? Or a new test in case it doesn't fit!

ljvmiranda921 (approved):

spacy_llm/tasks/textcat.py
```diff
@@ -137,10 +137,13 @@ def _format_response(self, response: str) -> Dict[str, float]:
         The returned dictionary contains the labels mapped to their score.
         """
         categories: Dict[str, float]
+        response = self._normalizer(response.strip())
```
kabirkhan, line 140: Should we call `response.strip()` here or rely on the normalizer to strip. 

I can't really think of a scenario where we want whitespace at the beginning or end of the response here if a user wanted to provide their own normalizer but I suppose it's possible.

spacy_llm/tasks/textcat.py
kabirkhan: you're right we should use the normalizer for sure for this. Made an update, let me know what you think to my question below.

spacy_llm/tasks/textcat.py
ljvmiranda921: I'm leaning towards delegating/relying it to the normalizer so that it's not "hidden" 🤔 and perhaps also for consistency (not too attached though, I agree we'd always want to strip())!

spacy_llm/tasks/textcat.py
svlandeg: IIRC, the normalizer is mainly meant to compare different labels, right? This is about stripping whitespace from the response, which feels like a natural thing to always do?

ljvmiranda921: @svlandeg If you put it that way I think that makes sense. @kabirkhan I'm ok reverting the commits from fc60358! 🙇

ljvmiranda921:
I reverted the last commit and ensured that the `response` is always `strip()`-ed. I think it's better if the `normalizer` is applied *only* on the `label`s :D Once @svlandeg approves, we can merge this! Thank you @kabirkhan for the PR 🙇

svlandeg (approved):
Thanks Kabir, good catch!

PR #80, opened 2023-05-12
Add optional logging of generated prompt/response

kabirkhan:
## Description

During development it can be really helpful to see the final prompt sent to the LLM Backend + the response generated. This adds a `verbose: bool` flag to the LLMWrapper component which will print the prompt + response when `__call__` is used.

This current PR is mostly to start a discussion. I'm not sold on adding a `print` to the component method.

Also, I'm happy to make this more generic and ensure this works with `pipe` but practically I don't think it's so useful to print out a ton of long generated prompts when you have multiple docs.

### Types of change

New Feature

## Checklist

- [ ] I ran the tests, and all new and existing tests passed.


spacy_llm/pipeline/llm.py
```diff
@@ -160,6 +166,7 @@ def __init__(
             binary .spacy files. Docs found in the cache directory won't be reprocessed.
         """
         self._name = name
+        self._task = task
```
rmitsch, line 169: I'm confused, why is this in this PR?

spacy_llm/pipeline/llm.py
```diff
@@ -178,12 +186,15 @@ def __call__(self, doc: Doc) -> Doc:
...
-            )
+            prompt = list(self._template([doc]))[0]
+            if self._verbose:
+                msg.divider("Generated Prompt")
+                print(prompt)  # noqa T201
+            response = list(self._backend([prompt]))[0]
+            if self._verbose:
+                msg.divider("Response from Backend")
+                print(response)  # noqa T201
+            docs = list(self._parse([doc], [response]))
```
rmitsch, line 197: - We should `assert` that the converted list has length 1.
- As you mentioned, `print()` isn't great. As `spacy-llm` should be integrated into spaCy soonish, let's use the same logging approach as there?
- Since we have to check for `self._verbose` everytime, it'd be nice to have a helper function `def log(verbose: bool, msg: str)` (maybe in a new `util.py`) that does that so we can log in a single line.

rmitsch (commented):
Thanks for the PR! I disagree on not logging in `.pipe()`. If users want to do that, we should let them (maybe they redirect output into a file and want to parse that file later?).

spacy_llm/pipeline/llm.py
kabirkhan: * For sure, I'll probably separate this into its own method that operates on batches of docs, then call that method in `__call__` and `pipe` both
* what is the spaCy logging approach? that's not so clear to me. 
* Sure that seems fine to me but want to make sure there's consistency with spaCy

spacy_llm/pipeline/llm.py
kabirkhan: mistake! I added this when I was debugging this change, didn't mean to keep it in here

spacy_llm/pipeline/llm.py
rmitsch: > - what is the spaCy logging approach? that's not so clear to me.

IMO using `wasabi` (see the [CLI functionality](https://github.com/explosion/spaCy/blob/master/spacy/cli/evaluate.py#L148)).

> - Sure that seems fine to me but want to make sure there's consistency with spaCy

I think whether we put this in a helper function or not is an implementation detail and doesn't matter that much w.r.t. consistency.

@adrianeboyd's opinion is appreciated here.

adrianeboyd: We'd use `wasabi` for CLI output, but not within pipeline components.

In v3 we tried switching to `logging` loggers in various places to mixed success. We had a lot of trouble getting messages filtered as intended so I remember going back to python warnings in a number of places. We still have logging for training, but basically nowhere else in the library. (I see a few traces that I think aren't intentional.)

To me this feels like it should be using logging and be behind a flag, so if it's noisy then users can easily disable all of it at once. (Maybe there are some obvious solutions that we missed to configuring logging before, but I just remember none of it working like we'd intended for warnings in particular.)

And I would add that I'm also not sure how I feel about adding `verbose` as a component-level config option.

adrianeboyd: So maybe just properly configured logging is what we actually want? I've just not had any success in the library context with this...

rmitsch: > I've just not had any success in the library context with this...

How so?

adrianeboyd: The main problems were related to warnings. We couldn't get warnings working the way we wanted and the combination of having both python warnings and logger warnings (where users couldn't really tell the difference) was not good.

If this is below `WARNING` maybe it's not as much of an issue? I'm not sure how configurable you would need this to be in the end. For the training loggers it makes sense to have a bit more structure, but I'm not sure inference loggers need much more than what's provided by `logging`?

spacy_llm/pipeline/llm.py
```diff
@@ -238,26 +236,44 @@ def pipe(self, stream: Iterable[Doc], *, batch_size: int = 128) -> Iterator[Doc]
...
+                len(noncached_doc_batch),
+            )
+
+        prompts = list(self._task.generate_prompts(noncached_doc_batch))
+        responses = list(self._backend(prompts))
+        for prompt, response, doc in zip(prompts, responses, noncached_doc_batch):
+            logger.debug("Generated prompt for doc: %s", doc.text)
+            logger.debug(prompt)
+            logger.debug("Backend response for doc: %s", doc.text)
+            logger.debug(response)
```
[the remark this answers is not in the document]
kabirkhan: @rmitsch @adrianeboyd

spacy_llm/tests/test_cache.py
```diff
@@ -74,7 +74,9 @@ def test_caching(use_pipe: bool) -> None:
         nlp_2 = _init_nlp(tmpdir)
         [nlp_2(text) for text in texts]
         cache = nlp_2.get_pipe("llm")._cache  # type: ignore
-        assert cache._stats["hit"] == n
+        # The number of hits should be batch_size n * 2 since
+        # __contains__ and __getitem__ both log hits
+        assert cache._stats["hit"] == n * 2
```
kabirkhan, line 79: Not sure why we weren't getting a failure here before. But this is technically correct based on the impl of the cache. I don't really think this is how the stats should work though. Shouldn't these stats only increment in `__getitem__`? It's not super important

spacy_llm/pipeline/llm.py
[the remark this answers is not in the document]
rmitsch: I think that's a decent compromise for now.

spacy_llm/pipeline/llm.py
```diff
@@ -238,26 +236,44 @@ def pipe(self, stream: Iterable[Doc], *, batch_size: int = 128) -> Iterator[Doc]
...
-                        yield doc
-                    else:
-                        doc = next(modified_docs)
-                        self._cache.add(doc)
-                        yield doc
+                yield from iter(self._process_docs(doc_batch))
             except Exception as e:
                 error_handler(self._name, self, doc_batch, e)
 
+    def _process_docs(self, docs: List[Doc]) -> List[Doc]:
```
rmitsch, line 244: Let's add a docstring here.

spacy_llm/tests/test_cache.py
rmitsch: That's because you're using one more `__contains__()` per doc in your rewrite than the original implementation (for `__call__()`), which changes the stats accordingly. As for how the stats should behave - how about we are more granular about this? We could count the number of hits in `__getitem__()` and those in `__contains__()` separately and call them `hits_getitem` and `hits_contains` (and maybe one `hits` counting all hits, regardless of how they happen).

spacy_llm/pipeline/llm.py
```diff
@@ -209,22 +215,18 @@ def __init__(
             max_batches_in_mem=int(cache["max_batches_in_mem"]),
             vocab=vocab,
         )
+        if not Doc.has_extension("llm_prompt"):
+            Doc.set_extension("llm_prompt", default=None)
+        if not Doc.has_extension("llm_response"):
+            Doc.set_extension("llm_response", default=None)
```
adrianeboyd, line 221: I would move this to `__call__` and `pipe` (repeated in both) so that it works with multiprocessing with spawn.

adrianeboyd: Or I guess possibly `_process_docs` instead, but it definitely needs to be called again in `pipe`.

spacy_llm/pipeline/llm.py
```diff
@@ -1,20 +1,26 @@
...
 
+
 from .. import registry  # noqa: F401
 from ..cache import Cache
 from ..compat import TypedDict
 from ..ty import LLMTask, PromptExecutor
 
 
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
```
adrianeboyd, line 21: And I'm also not sure what happens in terms of multiprocessing with this approach?

adrianeboyd:
Let's add some tests for mock logging and custom extensions with 1 and 2 processes (where you have to make sure to skip tests with 2 processes unless you're running on CPU)?

spacy_llm/pipeline/llm.py
kabirkhan: by multiprocessing do you just mean `nlp.pipe(..., n_processes=2)`?

adrianeboyd: Yes, with `nlp.pipe(n_process=1)` and `nlp.pipe(n_process=2)`, similar to the tests starting here: https://github.com/explosion/spaCy/blob/6930a6bf454bdae4ef12593c42896bf093315f8e/spacy/tests/test_language.py#L286

spacy_llm/pipeline/llm.py
kabirkhan: pulling this out into its own PR

spacy_llm/pipeline/llm.py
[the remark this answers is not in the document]
adrianeboyd: I think it might be confusing when the prompt and response values are separate logging calls from the header/doc info (and might end up out of order or interleaved)? I think it might be useful to combine the prompt / response into one line each?

spacy_llm/pipeline/llm.py
kabirkhan: Adding these kinds of tests separately in this PR: https://github.com/explosion/spacy-llm/pull/117. This works ok with n_process > 1 but we would need proper logging configuration in spaCy's multiprocessing stuff to make this work perfectly and to be able to follow the multiprocessing logging cookbook: https://docs.python.org/3/howto/logging-cookbook.html.

This logging to a `QueueHandler` idea would require some work on spaCy.

spacy_llm/cache.py
```diff
@@ -55,7 +56,7 @@ def _init_cache_index(self) -> None:
         if self._path is None:
             return
 
-        if self._path.exists() and not self._path.is_dir():
+        if self._path.exists() and not os.path.isdir(self._path):
```
bdura, line 59: Is that a Python 3.6 issue?

spacy_llm/pipeline/llm.py
```diff
@@ -237,28 +234,47 @@ def pipe(self, stream: Iterable[Doc], *, batch_size: int = 128) -> Iterator[Doc]
...
-                        yield doc
-                    else:
-                        doc = next(modified_docs)
-                        self._cache.add(doc)
-                        yield doc
+                yield from iter(self._process_docs(doc_batch))
             except Exception as e:
                 error_handler(self._name, self, doc_batch, e)
 
+    def _process_docs(self, docs: List[Doc]) -> List[Doc]:
```
bdura, line 242: I think this could technically be an `Iterable`?

```suggestion
    def _process_docs(self, docs: Iterable[Doc]) -> List[Doc]:
```

bdura (commented):
`test_llm.py` should be cleaned, otherwise everything looks good

rmitsch:
@kabirkhan What's the status on this, looks pretty much done to me?

README.md
```diff
@@ -200,6 +200,71 @@ labels = LABEL1,LABEL2,LABEL3
...
+Then when using the pipeline you'll be able to view the prompt and response.
+
+e.g. with the config and code from [Example 1](##example-1-add-a-text-classifier-using-a-gpt-3-model-from-openai) above
+
+
+```python
+from spacy import util
+
+config = util.load_config("config.cfg")
+nlp = util.load_model_from_config(config, auto_fill=True)
```
adrianeboyd, line 230: Update this to use `assemble`

kabirkhan:
@rmitsch I think this is ready to merge now after the update to using `assemble` in the docs

README.md
```diff
@@ -202,6 +202,70 @@ labels = LABEL1,LABEL2,LABEL3
 my_other_config_val = 0.3
 ```
 
+## Logging
+
+spacy-llm has a built-in logger that can log the actual prompt sent to the Backend LLM as well as the raw response sent back from the LLM. This logger uses the debug level and by default has a `logging.NullHandler()` configured.
```
rmitsch, line 207:
```suggestion
spacy-llm has a built-in logger that can log the prompt sent to the LLM as well as its raw response. This logger uses the debug level and by default has a `logging.NullHandler()` configured.
```

rmitsch: Maybe also describe what `logging.NullHandler()` does?

README.md
```diff
@@ -202,6 +202,70 @@ labels = LABEL1,LABEL2,LABEL3
...
+spacy_llm.logger.addHandler(logging.StreamHandler())
+spacy_llm.logger.setLevel(logging.DEBUG)
+```
+
+> NOTE: Any `logging` handler will work here so you probably want to use some sort of rotating `FileHandler` as the generated prompts can be quite long, especially for tasks with few-shot examples.
+
+
+Then when using the pipeline you'll be able to view the prompt and response.
+
+e.g. with the config and code from [Example 1](##example-1-add-a-text-classifier-using-a-gpt-3-model-from-openai) above
```
rmitsch, line 225:
```suggestion
E.g. with the config and code from [Example 1](##example-1-add-a-text-classifier-using-a-gpt-3-model-from-openai) above:
```

README.md
```diff
@@ -202,6 +202,70 @@ labels = LABEL1,LABEL2,LABEL3
...
+Text:
+'''
+You look gorgeous!
+'''
+
+Backend response for doc: You look gorgeous!
+COMPLIMENT
+```
+
+And the `print` of doc.cats to standard output should look like:
```
rmitsch, line 263:
```suggestion
`print(doc.cats)` to standard output should look like:
```

spacy_llm/pipeline/llm.py
```diff
@@ -148,8 +152,19 @@ def _process_docs(self, docs: List[Doc]) -> List[Doc]:
...
+                "Found %d docs in cache. Processing %d docs not found in cache",
+                len(docs) - len(noncached_doc_batch),
+                len(noncached_doc_batch),
+            )
+
+        prompts = list(self._task.generate_prompts(noncached_doc_batch))
+        responses = list(self._backend(prompts))
+        for prompt, response, doc in zip(prompts, responses, noncached_doc_batch):
+            logger.debug("Generated prompt for doc: %s\n%s", doc.text, prompt)
+            logger.debug("Backend response for doc: %s\n%s", doc.text, response)
```
rmitsch, line 166: We discussed internally we want to move away from the term `Backend` in user-facing docs and code, as it introduces unnecessary complication. (Stuff will mostly work the same, we'll just operate in terms of "models" instead of "backends" and focus on the model over the backend used to access it).
So let's use "LLM" or "Model" instead of "Backend" here.
```suggestion
            logger.debug("LLM response for doc: %s\n%s", doc.text, response)
```

rmitsch (commented):
Almost there!
- Left some minor comments w.r.t. docs and boilerplate output.
- Can you fix
  - the pre-commit issues?
  - the merge conflicts?

After resolving these things we're ready to merge.

README.md
kabirkhan: Wait shouldn't this be lowercase?

adrianeboyd: I agree with Raphael here.

rmitsch:
@kabirkhan Is this ready for a final review/merge? Please don't forget to move the task away from your column on the board, if so.

PR #81, opened 2023-05-12
Make a separate factory for parsing the input to the LLM task

kabirkhan:
## Description

Currently the actual Task classes are what gets registered. This leads to us parsing/editing the input to the class in the `__init__` method rather than relying on proper data. 

Separating this step out into a separate factory function, we can more easily support different inputs while retaining a clean interface for the task's `__init__`

This also adds support for defining inline examples and an inline custom template to show how that data can be parsed separately in a factory function rather than doing even more complex parsing of the config inputs in the init method.

### Types of change

New Feature/Enhancment

## Checklist

- [x] I confirm that I have the right to submit this contribution under the project's MIT license.
- [x] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #84, opened 2023-05-15
update NER prompt and add label definitions to better instruct the LLM

kabirkhan:
## Description

The existing NER prompt is a little brief and never explains to the model what each label actually means. Sure the language model has some context for the words in the label but in some experiments I've found that explicitly defining what we're looking for each label helps the LLM significantly. This also updates the general language of the NER prompt.

Previously, if I wanted the label `PER` (PERSON) to be extracted, then that's all the context the model would have (literally just the `PER` label). Now each label can be given a definition which is just some custom str input from the user when configuring the pipeline component.

This can cover more contexts than providing a few actual examples and I think it's a bit easier to reason about for a user. That being said, both definitions and examples can be provided and that's probably helpful. 

## Example definitions

let's say I want the labels PERSON and BASKETBALL_TEAM. I can create an LLM pipe with the following config:

```python
import spacy

nlp = spacy.blank("en")
nlp.add_pipe(
    "llm",
    name="llm_ner",
    config={
        "task" : {
            "@llm_tasks": "spacy.NER.v1",
            "labels": "PERSON,BASKETBALL_TEAM",
            "label_definitions": {
                "PERSON": "And named individual found in the text. e.g. Elmo, Kabir, Mr. Smith, Chewbacca",
                "BASKETBALL_TEAM": "Any names of basketball teams in the NBA. Do not include names of teams for any other sport."
            }
        },
        "backend": {
            "@llm_backends": "spacy.REST.v1",
            "api": "OpenAI",
            "config": {"model": "gpt-3.5-turbo"},
        },
    }
)

doc = nlp(
    "The Golden State Warriors beat the LA Lakers by almost 20 points on the back of an amazing performance from Steph Curry. "
    "In other sports news Man City was able to tie Real Madrid away from home with a stunning strike from Kevin De Bruyne"
)

print([(ent.text, ent.label_) for ent in doc.ents])

# [('Golden State Warriors', 'BASKETBALL_TEAM'),
# ('LA Lakers', 'BASKETBALL_TEAM'),
# ('Steph Curry', 'PERSON'),
# ('Kevin De Bruyne', 'PERSON')]
```

The model sometimes can separate out the names of basketball teams without these definitions but what if I wanted to extract something totally different or more specific. Just as an exercise, as a user I can define `BASKETBALL_TEAM` to actually mean "some number" by giving it a new definition.

e.g. 

```python
import spacy

nlp = spacy.blank("en")
nlp.add_pipe(
    "llm",
    name="llm_ner",
    config={
        "task" : {
            "label_definitions": {
                "BASKETBALL_TEAM": "While this label says BASKETBALL_TEAM. What you should really extract are all numbers, e.g. 1,30,100,30000, etc"
                 ...
            }
             ...
        },
        ...
)

## Calling with the same input now gives new entities based on the definition provided for the BASKETBALL_TEAM label
doc = nlp(
    "The Golden State Warriors beat the LA Lakers by almost 20 points on the back of an amazing performance from Steph Curry. "
    "In other sports news Man City was able to tie Real Madrid away from home with a stunning strike from Kevin De Bruyne"
)

print([(ent.text, ent.label_) for ent in doc.ents])

# ('20', 'BASKETBALL_TEAM'),
# ('Steph Curry', 'PERSON'),
# ('Kevin De Bruyne', 'PERSON')]
```

## Checklist


rmitsch:
Good idea! What's the reason this is still in draft?

kabirkhan:
Doesn't need to be in draft, just wondering if I should add a usage example for this to the docs.

rmitsch:
> Doesn't need to be in draft, just wondering if I should add a usage example for this to the docs.

Yes, please do.

spacy_llm/tasks/ner.py
```diff
@@ -59,34 +59,51 @@ def _unique(items: Iterable[str]) -> Iterable[str]:
 @registry.llm_tasks("spacy.NER.v1")
```
adrianeboyd, line 59: Since these changes are modifying the config, this should be become `spacy.NER.v2` and the package should maintain a copy of `spacy.NER.v1` with its original definition.

kabirkhan: hmm on principal I agree but this is an optional argument and the changing of the default template is internal. We could move this argument to the end of the `__init__` signature as well in case somebody was instantiating this task manually. It seems pretty low stakes to me to keep this as `spacy.NER.v1`. 

If we do make this `v2` then I assume we'd need a legacy package right?

adrianeboyd: The saved configs are always auto-filled, so it has a default but it's not exactly optional. The problem is that someone with a saved config that includes the new setting couldn't load their config with an older release of the library even when it looks like the exact same registered function is available.

So you always need to modify the registered function name if you're making any changes to the set of config settings. Whether you'd need to increment the version number for changes in behavior is a separate question. Here I might be surprised as a user if I upgraded the library and suddenly got wildly different results for my NER task because the prompt was changed so much in the background, so it might be better to increment the version even if the config settings were the same. For minor bug fixes or modifications in behavior we wouldn't always modify the name, though, but we'd only make this change on a minor version of the library. (This is also related to how `spacy package` fills in requirements by default, which is always one minor version range.)

You wouldn't necessarily need a legacy package. For now it could just be included here in parallel, similar to the registered functions in `spacy-transformers` or `spacy-loggers`.

kabirkhan: Ok great points! thanks! I've named this new task `v2` and have the original template in its own factory function. 
Before we actually release this v2 officially I'd like to at least have this PR merged as well: https://github.com/explosion/spacy-llm/pull/81 so we don't need to do a `v3` task.

spacy_llm/tasks/ner.py
```diff
@@ -92,12 +90,91 @@ class NERTask:
...
+):
+    task = NERTask(
+        labels=labels,
+        examples=examples,
+        normalizer=normalizer,
+        alignment_mode=alignment_mode,
+        case_sensitive_matching=case_sensitive_matching,
+        single_match=single_match,
+    )
+    setattr(task, "_TEMPLATE_STR", _NER_TEMPLATE_V1)
```
adrianeboyd, line 114: Why does this need to be `setattr`?

spacy_llm/tasks/ner.py
```diff
@@ -92,12 +90,91 @@ class NERTask:
...
+    return task
+
+
+@registry.llm_tasks("spacy.NER.v2")
```
adrianeboyd, line 118: Let's go ahead and move this into a separate method, too? I think it will be easier to maintain if you don't have constrain the `__init__` to correspond to the registered function.

spacy_llm/tasks/ner.py
adrianeboyd: And this feels like it ought to be an `NERTask` property instead maybe? (Where you'd set this from a template in `initialize` if you needed to? And where it should be saved with `llm.to_disk`? But that doesn't need to be in this PR.)

README.md
```diff
@@ -247,6 +247,7 @@ examples = null
 | Argument                  | Type                                    | Default      | Description                                                                                                                                  |
 | ------------------------- | --------------------------------------- | ------------ | -------------------------------------------------------------------------------------------------------------------------------------------- |
 | `labels`                  | `str`                                   |              | Comma-separated list of labels.                                                                                                              |
+| `label_descriptions`      | `Optional[Dict[str, str]]`              | `None`       | Optional dict of label: description of label to better instruct the LLM on what to extract for each label.                                                                                                              |
```
adrianeboyd, line 250: (plus reformatting)

```suggestion
| `label_descriptions`      | `Optional[Dict[str, str]]`              | `None`       | Optional dict of label descriptions: descriptions of labels to better instruct the LLM on what to extract for each label.                                                                                                              |
```

kabirkhan:
@adrianeboyd [these PRs](https://github.com/explosion/spacy-llm/pull/81) are pretty inter-related now. The goal for this PR was to be as minimal as possible and just update the prompt. That's why e.g. I didn't make the `spacy.NER.v2` decorator for a separate factory method but instead just on the class. It's also why we're doing that `setattr` of the template str. #81 handles this stuff and I didn't want to intermingle them too much

README.md
kabirkhan: Hmm I think the way it is now makes more sense? 

The dict will look like:

```
{
    "PERSON": "Any named individual in the text, e.g. ..."
}
```

So spelling out that it's a dict of `label: description of label` is correct

spacy_llm/tasks/ner.py
kabirkhan: Agreed, but see #81

spacy_llm/tasks/ner.py
kabirkhan: Agreed, but see #81

README.md
adrianeboyd: That is not clear at all from the current formatting.

spacy_llm/tasks/ner.py
adrianeboyd: But why isn't it just `task._TEMPLATE_STR = _NER_TEMPLATE_V1`?

spacy_llm/tasks/ner.py
adrianeboyd: Okay, that can wait then.

adrianeboyd:
I can see where it makes sense to postpone a few changes to #81. In this PR I still think that the README should be updated to use v2 and there should be explicit tests for both v1 and v2 (through a config / by registry name) in the test suite.

spacy_llm/tasks/ner.py
kabirkhan: oh pyright error in VSCode for me. But I just didn't really look at the error, fixed this

README.md
kabirkhan: Ok, I just rewrote this. How's the new definition?

rmitsch (commented):
I think this is almost there. 
- Merge conflicts have to be resolved.
- Shouldn't this be merged into #81 instead of into main?

rmitsch (approved):

PR #86, opened 2023-05-15
WIP: Explore using the PhraseMatcher for find_substrings

ljvmiranda921:
This PR attempts to rewrite the `find_substrings` function with the [PhraseMatcher](https://spacy.io/api/phrasematcher). Almost all of our options like case-sensitivity and matching can be replicated, except the `single_match` toggle. I implemented it naively by keeping track of seen `span.text`s and adding it to the final `offsets` output.

## Description

There are some behavioural changes with this approach. I made sure to document them in the tests below:

- The substrings will now appear as they're sorted by `PhraseMatcher`. At first, this causes some of our existing tests to fail. But I update them in ae49c04
- Partial substrings won't be matched anymore. Normally, a text that says:
 
```python
text = "Frodo Baggins and Samwise Gamgee went to Mt. Doom"
llm_predictions = "PER: Frodo B. Samwise G"
# Before
find_substrings_main_branch(text, substrings=llm_predictions)  # returns ["Frodo", "Samwise"]
# In this PR
find_substrings_this_branch(text, substrings=llm_predictions)  # returns []
```

Will be matched by using the `expand` option in the existing implementation. But in this case, the `PhraseMatcher` won't be able to find them. And it will return an empty list. You'll notice this behaviour in the **currently failing tests** in the CI. I didn't change the existing test here because I want to document that change in behaviour.

- The `spacy.NER.v1` task now takes in an optional `lang: str = "xx"` parameter that will be used to initialize the `PhraseMatcher`.

### Final thoughts

I don't mind sticking to the `find_substrings` option. I'm still looking for ways to solve the above issue.

### Types of change

- Refactor + Optimization
- Exploration
- New tests reflecting this update

## Checklist

- [ ] I ran the tests, and all new and existing tests passed.

### DIFF
--- a/spacy_llm/tasks/ner.py
+++ b/spacy_llm/tasks/ner.py
@@ -35,8 +35,8 @@ def find_substrings(
     matches: List[Span] = matcher(doc, as_spans=True)
 
     def _filter_first_hit(matches: List[Span]) -> Iterable[Tuple[int, int]]:
-        seen_spans = []
-        filtered = []
+        seen_spans: List[Span] = []
+        filtered: List[Span] = []
         for match in matches:
             start, end = match.start_char, match.end_char
             span_text = doc.char_span(start, end).text
--- a/spacy_llm/tasks/ner.py
+++ b/spacy_llm/tasks/ner.py
@@ -1,8 +1,10 @@
 from typing import Callable, Dict, Iterable, List, Optional, Tuple, Any
 
 import jinja2
+import spacy
 from pydantic import BaseModel
-from spacy.tokens import Doc
+from spacy.matcher import PhraseMatcher
+from spacy.tokens import Doc, Span
 from spacy.util import filter_spans
 
 from ..registry import lowercase_normalizer, registry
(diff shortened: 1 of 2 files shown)


rmitsch:
Do I understand correctly that the motivation for this PR is to use `PhraseMatcher` instead of custom code?

> Partial substrings won't be matched anymore. Normally, a text that says: ...

Yeah, that doesn't seem ideal. That's the issue you are referring to with

> I'm still looking for ways to solve the above issue.

Right?

ljvmiranda921:
Hi @rmitsch , yes! The motivation is to replace the `find_substrings` custom function with the `PhraseMatcher`. Stepping back, I think it's nice for reusability and it seems more efficient. 

Yeah, the partial substrings issue is something that I'm still working on. It works with our current code but I haven't replicated this capability yet with the `PhraseMatcher` (or any one of the other `Matcher`s)

spacy_llm/tasks/ner.py
```diff
@@ -18,41 +20,38 @@ def find_substrings(
...
-        while True:
-            start = text.find(substring, search_from)
-            if start == -1:
-                break
-            end = start + len(substring)
-            offsets.append((start, end))
-            if single_match:
-                break
-            search_from = end
+    nlp = spacy.blank(lang)
```
kabirkhan, line 29: I found myself wanting access the `nlp` (really the `vocab`) in the NER task as an input to the prompt. (e.g. explicitly saying which language the text is coming from). 

So rather than creating some other separate Language object here, let's pipe the vocab instance through and just use that here.

spacy_llm/tasks/ner.py
```diff
@@ -18,41 +20,38 @@ def find_substrings(
...
+            if not case_sensitive:
+                span_text = span_text.lower()
+            if span_text not in seen_spans:
+                filtered.append((start, end))
+                seen_spans.append(span_text)
+        return filtered
+
+    offsets = (
+        _filter_first_hit(matches)
+        if single_match
```
kabirkhan, line 52: what's the point of single_match again?

ljvmiranda921: To be honest, I copied this over way back from the original [explosion/prodigy-openai-recipes](https://github.com/explosion/prodigy-openai-recipes) repo. My best guess: there may be tokens that have the same `ORTH` but function differently in a text (hence different entities) which I've also seen quite often in our internal project. However, I don't think `single_match` solves this problem one-to-one.

spacy_llm/tasks/ner.py
```diff
@@ -18,41 +20,38 @@ def find_substrings(
...
+    matcher = PhraseMatcher(nlp.vocab, attr="ORTH" if case_sensitive else "LOWER")
+    patterns = [nlp.make_doc(substring) for substring in substrings]
+    matcher.add("substrings", patterns)
+    # Perform the match
+    doc = nlp.make_doc(text)
+    matches: List[Span] = matcher(doc, as_spans=True)
+
+    def _filter_first_hit(matches: List[Span]) -> Iterable[Tuple[int, int]]:
+        seen_spans = []
+        filtered = []
```
rmitsch, line 39:
```suggestion
        seen_spans: List[Span] = []
        filtered: List[Span] = []
```

rmitsch (commented):
This looks reasonable to me so far. I'll re-review after the partial string matching issue has been resolved.

ljvmiranda921:
Oks thanks for the review! I'm a bit stuck with the partial substring problem (will focus on it tomorrow). 
I feel like using the `Matcher` with some `REGEX` rules can be a bit hairy.

rmitsch:
Closing for now in favor of focusing on other issues.

PR #95, opened 2023-05-16
Add JinjaReader to read external template files

ljvmiranda921:
## Description

This Pull Request adds a `spacy.JinjaReader.v1` registered in the `miscs` registry so that we can pass external `.jinja2` templates in the files. The function signature returns a string instead of a `jinja2.Template`, so the "actual" creation of a template that can be rendered happens downstream. 

### Types of change

- New custom function

## Checklist


ljvmiranda921:
@rmitsch , I'm hitting a `FileNotFoundError` again just like the last time. Did we do something special to fix the problem with the `EXAMPLES_DIR`? I can't seem to surmise it in the codebase. 🤔

adrianeboyd:
There's a list of file endings in `MANIFEST.in` for files that should be included in the package.

README.md
```diff
@@ -258,6 +258,7 @@ examples = null
 | Argument                  | Type                                    | Default      | Description                                                                                                                                  |
 | ------------------------- | --------------------------------------- | ------------ | -------------------------------------------------------------------------------------------------------------------------------------------- |
 | `labels`                  | `str`                                   |              | Comma-separated list of labels.                                                                                                              |
+| `template`                | `Union[str, Callable[[], str]]`         |              | String or a function that returns the prompt template to render.                                                                             |
```
ljvmiranda921, line 261: Unsure of this. There is a default, actually, the `_DEFAULT_NER_TEMPLATE`. Not sure what to put in the `Default` column.

spacy_llm/tasks/ner.py
```diff
@@ -99,19 +99,24 @@ def _unique(items: Iterable[str]) -> Iterable[str]:
...
     case_sensitive_matching: bool = False,
     single_match: bool = False,
 ) -> "NERTask":
     labels_list = split_labels(labels)
+    # Load examples
     raw_examples = examples() if callable(examples) else examples
     ner_examples = [NERExample(**eg) for eg in raw_examples] if raw_examples else None
+    # Load templates
+    if template is None:
+        raise ValueError("A template must be supplied. It cannot be 'None'.")
```
ljvmiranda921, line 115: Decided to just raise an error when `template` is `None` rather than assigning default. Not too attached with this decision though. Happy to hear alternatives!

ljvmiranda921:
Ready for review! Note that this is dependent on #81 (it merges to that branch) cc: @rmitsch @kabirkhan

README.md
rmitsch: Hm, we could move _DEFAULT_NER_TEMPLATE into a registry function and specify the registry handle of that function here? Although it seems odd to do this just for the docs. Alternatively a "See here" linking to `_DEFAULT_NER_TEMPLATE` in the code?

README.md
```diff
@@ -315,6 +316,7 @@ examples = null
 | Argument            | Type                                    | Default | Description                                                                                                                                      |
 | ------------------- | --------------------------------------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
 | `labels`            | str                                     |         | Comma-separated list of labels.                                                                                                                  |
+| `template`          | `Union[str, Callable[[], str]]`         |         | String or a function that returns the prompt template to render.                                                                                 |
```
rmitsch, line 319: Same w.r.t. default value as above.

rmitsch (commented):
Looks good! There'll be more things to test in terms of path resolution and (de-)serialization, but I think this is fine for now. Just minor comments.

I'll merge this after https://github.com/explosion/spacy-llm/pull/81 to keep the diff there clean.

README.md
ljvmiranda921: I like the latter suggestion!

ljvmiranda921:
Thanks for the reviews! I'll fix the conflicts once #81 's been merged! 🙇 (so it's just a one time conflict fix)

spacy_llm/registry/reader.py
```diff
@@ -6,6 +6,23 @@
 from .util import registry
 
 
+@registry.misc("spacy.JinjaReader.v1")
```
adrianeboyd, line 9: I feel like this has moved away from how it was originally envisioned and should  probably be turned into a more generic file reader instead? All it does is require a particular file ending and then read in a string.

And in terms of the file extensions it seems overly restrictive for jinja where there are multiple file endings in common use? (spacy uses `.jinja`)

spacy_llm/ty.py
```diff
@@ -12,6 +12,7 @@
 ExamplesConfigType = Union[
     Iterable[Dict[str, Any]], Callable[[], Iterable[Dict[str, Any]]], None
 ]
+TemplateConfigType = Union[str, Callable[[], str]]
```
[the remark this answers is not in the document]
ljvmiranda921: Oh I admit I was just copying the `ExamplesConfigType` here. I thought this means: accept any string and a `JinjaReader` (which is a `Callable` that returns a string). Sometimes, I'm still confused by types so let me know what's the appropriate thing to do!

spacy_llm/tasks/ner.py
```diff
@@ -53,9 +53,12 @@ def make_ner_task_v2(
     labels_list = split_labels(labels)
     raw_examples = examples() if callable(examples) else examples
     span_examples = [SpanExample(**eg) for eg in raw_examples] if raw_examples else None
+    if template is None:
+        raise ValueError("A template must be supplied. It cannot be 'None'.")
+    task_template = template() if callable(template) else template
```
rmitsch, line 58: IMO `template` should be resolved by the config system (if e. g. `spacy.FileReader.v1` is set), so I don't think we ever get a `Callable` here?

spacy_llm/tests/tasks/test_textcat.py
```diff
@@ -131,9 +173,12 @@ def test_textcat_config(task, cfg_string, request):
...
 @pytest.mark.skipif(has_openai_key is False, reason="OpenAI API key not available")
 @pytest.mark.parametrize("task", ["binary", "multilabel_nonexcl", "multilabel_excl"])
-@pytest.mark.parametrize("cfg_string", ["zeroshot_cfg_string", "fewshot_cfg_string"])
+@pytest.mark.parametrize(
+    "cfg_string",
+    ["zeroshot_cfg_string", "fewshot_cfg_string", "ext_template_cfg_string"],
+)
 def test_textcat_predict(task, cfg_string, request):
-    """Use OpenAI to get zero-shot Textcat results
+    """Use OpenAI to get Textcat results
```
rmitsch, line 181:
```suggestion
    """Use OpenAI to get Textcat results.
```

spacy_llm/tests/tasks/test_textcat.py
```diff
@@ -131,9 +173,12 @@ def test_textcat_config(task, cfg_string, request):
...
     Note that this test may fail randomly, as the LLM's output is unguaranteed
     to be consistent/predictable
```
rmitsch, line 183:
```suggestion
    to be consistent/predictable.
```

spacy_llm/registry/reader.py
```diff
@@ -6,6 +6,21 @@
 from .util import registry
 
 
+@registry.misc("spacy.FileReader.v1")
+def file_reader(path: Union[str, Path]) -> str:
+    """Read a file from a path and return its contents
```
rmitsch, line 11:
```suggestion
    """Read a file from a path and return its contents.
```

spacy_llm/tasks/rel.py
```diff
@@ -68,9 +68,12 @@ def make_rel_task(
     labels_list = split_labels(labels)
     raw_examples = examples() if callable(examples) else examples
     rel_examples = [RELExample(**eg) for eg in raw_examples] if raw_examples else None
+    if template is None:
+        raise ValueError("A template must be supplied. It cannot be 'None'.")
+    task_template = template() if callable(template) else template
```
rmitsch, line 73: Same here.

spacy_llm/tasks/spancat.py
```diff
@@ -52,9 +52,12 @@ def make_spancat_task_v2(
     labels_list = split_labels(labels)
     raw_examples = examples() if callable(examples) else examples
     span_examples = [SpanExample(**eg) for eg in raw_examples] if raw_examples else None
+    if template is None:
+        raise ValueError("A template must be supplied. It cannot be 'None'.")
+    task_template = template() if callable(template) else template
```
rmitsch, line 57: Same here.

rmitsch (commented):
Almost good to go.

spacy_llm/tasks/ner.py
kabirkhan: fixing this

kabirkhan: Ok I removed this and also just remove the error if template is None. We have type hints here for a reason + we have a default template for each task so the user would need to explicitly set the template to None

ljvmiranda921:
From what I can tell from the diffs the big change here was removing the `Callable` and just using `str`. I think it looks good! Ok for merging later :slightly_smiling_face:

rmitsch (approved):

PR #96, opened 2023-05-16
Fix OpenAI requests for chat endpoint

rmitsch:
## Description

OpenAI's `/chat/completions/` endpoint doesn't support batching, so the previous approach to executing prompts batched for chat models was flawed and failed the external test workflow. This PR fixes the behavior by executing prompts individually, which unfortunately seems like the only* clean way to deal with this now.

\* One alternative way is modifying the prompt, but this seems inelegant and error-prone.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy_llm/backends/rest/backend/openai.py
+++ b/spacy_llm/backends/rest/backend/openai.py
@@ -1,6 +1,6 @@
 import os
 from enum import Enum
-from typing import Dict, Iterable, List, Union
+from typing import Dict, Iterable, List, Any, Sized
 
 import requests  # type: ignore[import]
 import srsly  # type: ignore[import]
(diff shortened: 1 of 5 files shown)


spacy_llm/backends/rest/backend/openai.py
```diff
@@ -97,42 +97,54 @@ def __call__(self, prompts: Iterable[str]) -> Iterable[str]:
         }
         api_responses: List[str] = []
         prompts = list(prompts)
-
-        data: Dict[str, Union[List[str], List[Dict[str, str]]]] = {}
         url = self._url if self._url else self.supported_models[self._config["model"]]
```
bdura, line 100: This is neither here nor there, but you could do:

```python
url = self._url or self.supported_models[self._config["model"]]
```

bdura (approved):
Looks good to me!

bdura (approved):
Looks good to me!

spacy_llm/backends/rest/backend/openai.py
rmitsch: Your suggestion is more elegant, I suspect it might also be a bit less clear? I personally prefer it, but I've had negative feedback in the past about being too clever with conditionals.

Not changing this now, but I'll keep it in mind!

PR #97, opened 2023-05-16
Extend test suite to contain all usage and readme examples

rmitsch:
## Description

Extend test suite to contain all usage and readme examples.

## Checklist

- [ ] I ran the tests (including the **external** tests), and all new and existing tests passed.

### DIFF
--- a/usage_examples/tests/conftest.py
+++ b/usage_examples/tests/conftest.py
@@ -1,0 +1,5 @@
+from spacy_llm.tests.conftest import (
+    pytest_addoption,  # noqa: F401
+    pytest_runtest_setup,  # noqa: F401
+    pytest_collection_modifyitems,  # noqa: F401
+)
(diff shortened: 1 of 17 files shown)


usage_examples/tests/conftest.py
```diff
@@ -0,0 +1,5 @@
+from spacy_llm.tests.conftest import (
+    pytest_addoption,  # noqa: F401
```
bdura, line 2: I usually add something like:

```toml
[tool.ruff.per-file-ignores]
# Ignore unused imports in __init__ files
"__init__.py" = ["F401"]
```

to my `pyproject.toml` file, that way I don't have to repeat it every time.

bdura (approved):
Looks great!

usage_examples/tests/conftest.py
rmitsch: That makes sense! Didn't know this is possible. In general I'd like to have all unused imports flagged though, so I won't add this for now. If this is a frequently occurring situation, I'll update the `ruff` config accordingly.

PR #98, opened 2023-05-16
Use NoOp REST backend in tests where reasonable

rmitsch:
## Description

Use NoOp REST backend in tests where reasonable to avoid unnecessary API calls.

### Types of change

Chore.

## Checklist

### DIFF
--- a/spacy_llm/tests/pipeline/test_llm.py
+++ b/spacy_llm/tests/pipeline/test_llm.py
@@ -69,7 +69,7 @@ def test_llm_serialize_disk():
 
 
 @pytest.mark.external
-def test_type_checking_valid(config) -> None:
+def test_type_checking_valid(noop_config) -> None:
     """Test type checking for consistency between functions."""
     # Ensure default config doesn't raise warnings.
     nlp = spacy.blank("en")
--- a/spacy_llm/tests/pipeline/test_llm.py
+++ b/spacy_llm/tests/pipeline/test_llm.py
@@ -1,5 +1,5 @@
 import warnings
-from typing import Iterable
+from typing import Iterable, Dict, Any
 
 import pytest
 import spacy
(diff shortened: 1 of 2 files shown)


spacy_llm/tests/pipeline/test_llm.py
```diff
@@ -72,7 +78,7 @@ def test_type_checking_valid() -> None:
         nlp.add_pipe("llm", config={"task": {"@llm_tasks": "spacy.NoOp.v1"}})
```
bdura, line 78: Should we be using the fixture in here?

bdura (commented):
I left two comments that may reflect the fact that I'm still new to this project. Otherwise , looks good.

spacy_llm/tests/pipeline/test_llm.py
```diff
@@ -62,8 +68,8 @@ def test_llm_serialize_disk():
         llm.from_disk(tmp_dir / "llm")
 
 
-@pytest.mark.skipif(has_openai_key is False, reason="OpenAI API key not available")
-def test_type_checking_valid() -> None:
+@pytest.mark.external
```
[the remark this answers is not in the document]
rmitsch: In this case it's unfortunately still necessary, because the OpenAI REST backend calls into the API at init time to ensure that the authentication works (and the OpenAI REST backend is the default config we're testing here).

spacy_llm/tests/pipeline/test_llm.py
rmitsch: No, we deliberately test the default config here. The fixture is for a NoOp backend setup.

spacy_llm/tests/pipeline/test_llm.py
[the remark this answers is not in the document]
bdura: Right. Thanks.

spacy_llm/tests/pipeline/test_llm.py
bdura: Sounds good. Then we don't need to pass the fixture though, but I think we're fine as it is.

bdura (approved):
Looks good!

PR #99, opened 2023-05-16
improve textcat task by better constraining its output and giving a more detailed task intro

kabirkhan:
## Description

The switch to `gpt-3.5-turbo` as the default model seems to make the LLM lean toward providing a longer output (not necessarily constrained to a single label).

e.g. 

for the text:

```
"you are quite beautiful"
```

the LLM response from `gpt-3.5-turbo` for binary classification was:

```
COMPLIMENT (POS)
```

The LLM got this right but not in a way we can parse. 

This PR adds a longer intro, similar to the new NER intro in: #84 as well as language to tell the model it's important not to return anything else but the labels we want for each kind of classification.

### Types of change

fix/enhancement

## Checklist

### DIFF
--- a/spacy_llm/tasks/textcat.py
+++ b/spacy_llm/tasks/textcat.py
@@ -16,18 +16,25 @@ class TextCatExample(BaseModel):
 @registry.llm_tasks("spacy.TextCat.v1")
 class TextCatTask:
     _TEMPLATE_STR = """
+You are an expert Text Classification system. Your task is to accept Text as input
+and provide a category for the text based on the predefined labels.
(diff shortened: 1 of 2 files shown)


rmitsch (approved):

rmitsch:
Did you run the external tests? Is this ready to be merged?

kabirkhan:
yep external tests are passing locally but are we not running those in CI?

rmitsch:
> yep external tests are passing locally but are we not running those in CI?

Only once a day (or triggered manually) to avoid incurring unnecessarily high costs.

PR #102, opened 2023-05-17
Add streamlit demo

victorialslocum:
## Description

Add a streamlit demo to show `spacy-llm` with OpenAI for NER and Textcat. 

### Types of change

enhancement to usage examples

## Checklist


usage_examples/streamlit/streamlit_app.py
```diff
@@ -0,0 +1,32 @@
+from spacy import util
+from spacy_streamlit import visualize_ner, visualize_textcat
+import streamlit as st
+
+DEFAULT_TEXT = "Matt Honnibal is the CEO of Explosion."
```
adrianeboyd, line 5: Let's pick a different example sentence?

rmitsch:
Note: the tests are failing because sharing env secrets with forks doesn't work (without compromising security to some degree). If you set an `OPENAI_API_KEY` env secret in your fork repo, this should be fixed.

usage_examples/streamlit/streamlit_app.py
```diff
@@ -0,0 +1,32 @@
...
+ner_config = util.load_config("usage_examples/streamlit/openai_ner_zeroshot.cfg")
+ner_model = util.load_model_from_config(ner_config, auto_fill=True)
+
+models = {"textcat": textcat_model, "ner": ner_model}
+model_names = models.keys()
+
+selected_model = st.sidebar.selectbox("Model", model_names)
+
+nlp = models[selected_model]
+doc = nlp(text)
```
ljvmiranda921, line 24: I'm wondering how streamlit's reactive process works with this. Does it call `nlp` as you're typing? If yes, it might be costly and it might be better to have this line be enclosed in some "press button" action.

ines: I think the default Streamlit text input field needs you to submit before it executes further!

usage_examples/streamlit/streamlit_app.py
```diff
@@ -0,0 +1,32 @@
+from spacy import util
+from spacy_streamlit import visualize_ner, visualize_textcat
+import streamlit as st
+
+DEFAULT_TEXT = "Matt Honnibal is the CEO of Explosion."
+
+st.title("spacy-llm Streamlit Demo")
```
ines, line 7: Maybe we can add a little intro here with a link to the repo and a short summary of what it is and what the demo does?

usage_examples/streamlit/streamlit_app.py
victorialslocum: Yep, exactly, you need to submit it first 😄

usage_examples/streamlit/streamlit_app.py
victorialslocum: Is the one I changed it to better? Suggestions are welcome

usage_examples/streamlit/streamlit_app.py
victorialslocum: Is [this](https://github.com/explosion/spacy-llm/blob/2cd47c01e95e2b774042efb655efe7160d539205/usage_examples/streamlit/streamlit_app.py#L11) enough, or should I expand upon anything?

victorialslocum:
> Note: the tests are failing because sharing env secrets with forks doesn't work (without compromising security to some degree). If you set an `OPENAI_API_KEY` env secret in your fork repo, this should be fixed.

Setting the env secret didn't seem to work, is there another way?

adrianeboyd:
I think we just agree to ignore the CI here for now because it's not doing anything useful? (Otherwise one possibility is to recreate the PR with a branch in the main repo.)

usage_examples/streamlit/streamlit_app.py
adrianeboyd: Better! (Still slightly old-white-dude for my taste, but it's fine if it works for both the demo cases.)

usage_examples/streamlit/streamlit_app.py
```diff
@@ -0,0 +1,46 @@
...
+    into robust outputs for various NLP tasks, no training data required.
+
+    This demo uses the OpenAI backend to demonstrate the NER and textcat 
+    tasks.
+    """
+)
+os.environ["OPENAI_API_KEY"] = st.text_input("Your OpenAI API key", type="password")
+text = st.text_area("Text to analyze", DEFAULT_TEXT, height=70)
+
+if os.environ["OPENAI_API_KEY"]:
```
adrianeboyd, line 23: I think this might crash if the key isn't set? (I can't remember exactly how streamlit works here.) Maybe `if "OPENAI_API_KEY" in os.environ:` ... `else` helpful-message?

victorialslocum: I tested it, it won't run anything in the `if` statement until the key is set, so I think it works?

adrianeboyd: I tried it with one of the `spacy-streamlit` demos and it gives a pretty ugly `KeyError` traceback. I think we can check for the key and provide a clearer error message if it's not set?

victorialslocum: Updated!

usage_examples/streamlit/streamlit_app.py
```diff
@@ -0,0 +1,48 @@
...
+    prompt = "\n".join([str(prompt) for prompt in nlp.get_pipe("llm")._template([doc])])
+    st.sidebar.markdown("## OpenAI Prompt:")
+    st.sidebar.text(prompt)
+
+    if selected_model == "textcat":
+        visualize_textcat(doc)
+    if selected_model == "ner":
+        visualize_ner(doc)
+else:
+    st.error("Input your OpenAI API key")
```
adrianeboyd, line 48: "To run this demo, set your OpenAI API key as the environment variable `OPENAI_API_KEY`"?

adrianeboyd: (Is there a normal way to print errors in streamlit?)

adrianeboyd: Ah, you're using it, nevermind!

victorialslocum: What about, "To run this demo, set your OpenAI API key as the environment variable OPENAI_API_KEY or input above"? We have the password field for it within the demo, I think it might be confusing to not address that?

usage_examples/streamlit/streamlit_app.py
adrianeboyd: Ah, I was pretty confused here about how you set this up, sorry. I guess then you always have `os.environ["OPENAI_API_KEY"]`, but it might be `""` or `None` and that's what your `if` is checking for?

victorialslocum: Yes I think so? I'm not entirely sure though. I think the `else` statement is good anyways.

adrianeboyd: I think the `else` statement is good.

Can you use the current `OPENAI_API_KEY` as the default value for this field, something like `os.environ.get("OPENAI_API_KEY", "")`?

usage_examples/streamlit/streamlit_app.py
adrianeboyd: Yeah, I was confused. I think your original proposal was fine.

usage_examples/streamlit/streamlit_app.py
victorialslocum: Fixed! Anything else?

usage_examples/streamlit/streamlit_app.py
```diff
@@ -0,0 +1,49 @@
...
+text = st.text_area("Text to analyze", DEFAULT_TEXT, height=70)
+
+if os.environ.get("OPENAI_API_KEY"):
+    textcat_config = util.load_config(
+        "usage_examples/streamlit/openai_textcat_zeroshot.cfg"
+    )
+    textcat_model = util.load_model_from_config(textcat_config, auto_fill=True)
+
+    ner_config = util.load_config("usage_examples/streamlit/openai_ner_zeroshot.cfg")
+    ner_model = util.load_model_from_config(ner_config, auto_fill=True)
```
adrianeboyd, line 31: Let's switch to `spacy_llm.util.assemble` here too (you'll still need the import further up).

```suggestion
    textcat_model = assemble(textcat_config)
    ner_model = assemble(ner_config)
```

adrianeboyd:
To go along with v0.2.0 I think it might be nice if this uses the v2 tasks so we can point to it as a demo for the updated prompts right away?

usage_examples/streamlit/streamlit_app.py
```diff
@@ -0,0 +1,49 @@
...
+
+    models = {"textcat": textcat_model, "ner": ner_model}
+    model_names = models.keys()
+
+    selected_model = st.sidebar.selectbox("Model", model_names)
+
+    nlp = models[selected_model]
+    doc = nlp(text)
+    prompt = "\n".join([str(prompt) for prompt in nlp.get_pipe("llm")._template([doc])])
```
adrianeboyd, line 40: It looks like this API has already changed. Is it possible to do this without `_template`?

usage_examples/streamlit/streamlit_app.py
adrianeboyd: Then you'd have:

```suggestion
    textcat_config = util.load_config(TEXTCAT_CONFIG)
    textcat_model = assemble_from_config(textcat_config)

    ner_config = util.load_config(NER_CONFIG)
    ner_model = assemble_from_config(ner_config)
```

usage_examples/streamlit/streamlit_app.py
```diff
@@ -0,0 +1,49 @@
...
+    Large Language Models (LLMs) into spaCy, featuring a modular system 
+    for fast prototyping and prompting, and turning unstructured responses 
+    into robust outputs for various NLP tasks, no training data required.
+
+    This demo uses the OpenAI backend to demonstrate the NER and textcat 
+    tasks.
+    """
+)
+
+os.environ["OPENAI_API_KEY"] = st.text_input("Your OpenAI API key", type="password")
```
adrianeboyd, line 21:
```suggestion
os.environ["OPENAI_API_KEY"] = st.text_input("Your OpenAI API key", type="password", value=os.environ.get("OPENAI_API_KEY", ""))
```

adrianeboyd (commented):
This is what I meant by filling in the default value for the openai key. Otherwise you're overwriting the existing key every time and the user would have to fill it in.

usage_examples/streamlit/streamlit_app.py
victorialslocum: Oh this makes so much more sense. Thanks!

usage_examples/streamlit/streamlit_app.py
victorialslocum: I can update this for v2!

adrianeboyd:
And I think it would make sense to move the prompt to the main column because it's too hard to read in the sidebar. Maybe below the visualization?

usage_examples/streamlit/streamlit_app.py
victorialslocum: I think `util.load_config` expects a `file` type input? Not a string?

adrianeboyd: Ah, the `load_config_from_str` instead.

victorialslocum:
Should I add the REL and Spancat tasks added in v0.2.0 as well? Or do we just want to keep it with NER and Textcat for now?

rmitsch:
> Should I add the REL and Spancat tasks added in v0.2.0 as well? Or do we just want to keep it with NER and Textcat for now?

If it's quick to add them, yes. Otherwise I'd say let's prioritize getting this PR done (we can still open a follow-up PR with more tasks).

victorialslocum:
All the feedback has been addressed, so unless there's anything more, I think it makes sense to merge this one and make updates as needed.

PR #104, opened 2023-05-17
fix: caching and cache interaction

bdura:
## Description

This PR documents and fixes two separate bugs involving the cache:

1. Documents were never actually registered within the cache (`_doc2batch` was never updated)
2. The `LLMWraper.pipe` method didn't iterate over the right documents.

### Types of change

This is a bug fix.

## Checklist


rmitsch (approved):
LGTM, thanks for the fix!

PR #107, opened 2023-05-18
fix response of openai chat endpoint

kabirkhan:
## Description

PR https://github.com/explosion/spacy-llm/pull/96 introduced a bug that wasn't caught in our tests due to using a Noop Task.

[This line](https://github.com/explosion/spacy-llm/blob/8279f6764cb56a0d0f6a809e1c436856c939ed0b/spacy_llm/backends/rest/backend/openai.py#LL135C5-L135C5) overwrites the `api_responses` variable so if multiple prompts are sent, only 1 response is returned (the last in the list). 

Fixing by using `append` and using the `TextCat` task instead of Noop so the docs/responses are actually used and tested.

### Types of change

fix

## Checklist


rmitsch (approved):
Good catch!

PR #110, opened 2023-05-19
Improve error handling in OpenAI REST backend

rmitsch:
## Description

Improve error handling in OpenAI REST backend: actual error message prefixes the stack trace as separate `ValueError`.

### Types of change

Enhancement.

## Checklist

  - [ ] `tests`
  - [x] `usage_example/tests`

### DIFF
--- a/spacy_llm/backends/rest/backend/openai.py
+++ b/spacy_llm/backends/rest/backend/openai.py
@@ -50,6 +50,13 @@ class OpenAIBackend(Backend):
                 "an environment variable 'OPENAI_API_KEY."
             )
 
+        # Ensure endpoint is supported.
+        url = self._url if self._url else self.supported_models[self._config["model"]]
+        if url not in (Endpoints.non_chat, Endpoints.chat):
+            raise ValueError(
+                f"Endpoint {url} isn't supported. Please use of: {Endpoints.chat}, {Endpoints.non_chat}."
+            )
+
         # Check the access and get a list of available models to verify the model argument (if not None)
         # Even if the model is None, this call is used as a healthcheck to verify access.
         headers = {
@@ -108,7 +115,14 @@ class OpenAIBackend(Backend):
                     timeout=self._timeout,
                 ),
             )
-            r.raise_for_status()
+            try:
+                r.raise_for_status()
+            except Exception as ex:
+                res_content = srsly.json_loads(r.content.decode("utf-8"))
+                # Include specific error message in exception.
+                raise ValueError(
+                    f"Request to OpenAI API failed: {res_content.get('error', {}).get('message', str(res_content))}"
+                ) from ex
             responses = r.json()
 
             if "error" in responses:
--- a/spacy_llm/tasks/noop.py
+++ b/spacy_llm/tasks/noop.py
@@ -7,16 +7,8 @@ from ..registry import registry
 
 @registry.llm_tasks("spacy.NoOp.v1")
 class NoopTask:
-    def __init__(self):
-        """Returns (1) templating Callable (injecting Doc data into a prompt template and returning one fully specified
-        prompt per passed Doc instance) and (2) parsing callable (parsing LLM responses and updating Doc instances with the
-        extracted information).
-        RETURNS (Tuple[Callable[[Iterable[Doc]], Iterable[str]], Any]): templating Callable, parsing Callable.
-        """
-        pass
-
     def generate_prompts(self, docs: Iterable[Doc]) -> Iterable[str]:
-        for doc in docs:
+        for _ in docs:
             yield "Don't do anything."
 
     def parse_responses(
--- a/spacy_llm/tests/backends/test_rest.py
+++ b/spacy_llm/tests/backends/test_rest.py
@@ -1,10 +1,14 @@
 # mypy: ignore-errors
 import copy
+import re
+from typing import Iterable
 
 import pytest
 import spacy
+from spacy.tokens import Doc
 
 from ..compat import has_openai_key
+from ...registry import registry
 
 PIPE_CFG = {
     "backend": {
@@ -27,8 +31,8 @@ def test_initialization():
 
 
 @pytest.mark.external
-def test_rest_backend_error_handling():
-    """Test error handling for default/minimal REST backend."""
+def test_model_error_handling():
+    """Test error handling for wrong model."""
     nlp = spacy.blank("en")
     with pytest.raises(ValueError) as err:
         nlp.add_pipe(
@@ -39,6 +43,41 @@ def test_rest_backend_error_handling():
             },
         )
     assert "The specified model 'x-gpt-3.5-turbo' is not available." in str(err.value)
+
+
+@pytest.mark.external
+def test_doc_length_error_handling():
+    """Test error handling for wrong URL."""
+    nlp = spacy.blank("en")
+
+    @registry.llm_tasks("spacy.Count.v1")
+    class CountTask:
+        def generate_prompts(self, docs: Iterable[Doc]) -> Iterable[str]:
+            for doc in docs:
+                yield f"Count the number of characters in this string: '{doc.text}'."
+
+        def parse_responses(
+            self, docs: Iterable[Doc], responses: Iterable[str]
+        ) -> Iterable[Doc]:
+            return docs
+
+    nlp.add_pipe(
+        "llm",
+        config={
+            # Not using the NoOp task is necessary here, as the NoOp task sends a fixed-size prompt.
+            "task": {"@llm_tasks": "spacy.Count.v1"},
+            "backend": {"config": {"model": "ada"}},
+        },
+    )
+    # Call with an overly long document to elicit error.
+    with pytest.raises(
+        ValueError,
+        match=re.escape(
+            "Request to OpenAI API failed: This model's maximum context length is 2049 tokens, however you requested "
+            "2527 tokens (2511 in your prompt; 16 for the completion). Please reduce your prompt; or completion length."
+        ),
+    ):
+        nlp("n" * 5000)
 
 
 @pytest.mark.parametrize("model", ("gpt-3.5-turbo", "text-davinci-002"))


spacy_llm/backends/rest/registry.py
```diff
@@ -12,7 +12,8 @@ def backend_rest(
     config: Dict[Any, Any] = SimpleFrozenDict(),
     strict: bool = True,
     max_tries: int = 3,
-    timeout: int = 30,
+    interval: int = 30,
```
kabirkhan, line 15: interval should probably start smaller since we're doubling it each time, ideally start at like 1 second and make max_tries a little higher (5 maybe)

requirements-dev.txt
```diff
@@ -7,7 +7,7 @@ mypy>=0.990,<1.1.0; platform_machine != "aarch64" and python_version >= "3.7"
 black==22.3.0
 types-requests==2.28.11.16
 # Prompting libraries needed for testing
-langchain>=0.0.144,<0.1; python_version>="3.9"
+langchain>=0.0.144,<0.0.174; python_version>="3.9"
```
rmitsch, line 10: Lowered the `langchain` pin because there seems to be an issue with their `sqlalchemy` usage on Python 3.11 since their v0.0.174 (released three days ago). We can revisit this in a bit, I expect they'll fix this soon.

spacy_llm/backends/rest/backend/base.py
```diff
@@ -38,7 +42,8 @@ def __init__(
         self._check_api_endpoint_compatibility()
 
         assert self._max_tries >= 1
-        assert self._timeout >= 1
+        assert self._interval >= 1
```
bdura, line 45: That seems a bit over-constrained? We should at least document it in the docstrings?

bdura: Ah those are integers... Is there a particular why? I think users would expects floats there

spacy_llm/backends/rest/backend/base.py
```diff
@@ -69,31 +74,46 @@ def credentials(self) -> Dict[str, str]:
...
+            response = _call_api()
             i += 1
-            timeout = timeout * 2  # Increase timeout everytime you retry
+            # Increase timeout everytime you retry
+            interval = interval * 2
 
         if response.status_code in self._retry_error_codes:
             raise ConnectionError(
-                f"API could not be reached within {self._timeout} seconds in {self._max_tries} attempts. Check your "
+                f"API could not be reached within {self._interval} seconds in {self._max_tries} attempts. Check your "
```
bdura, line 116: I'm wondering if there is a way we can reflect the fact that there was an exponential back-off? Could we document it in the docstrings? (I'll add a suggestion)

spacy_llm/backends/rest/backend/base.py
```diff
@@ -24,12 +26,14 @@ def __init__(
             be raised.
             Note that only response object structure will be checked, not the prompt response text per se.
         max_tries (int): Max. number of tries for API request.
-        timeout (int): Timeout for API request in seconds.
+        interval (int): Time interval (in seconds) for API retries in seconds.
```
bdura, line 29:
```suggestion
        interval (int): Time interval (in seconds) for API retries in seconds. We implement a base 2 exponential backoff at each retry.
```

The wording is far from perfect, but that seems like relevant information?

By the way, we could also accept a [Thinc schedule](https://thinc.ai/docs/api-schedules#compounding) (most probably overkill, but still)

spacy_llm/backends/rest/backend/base.py
```diff
@@ -69,31 +74,46 @@ def credentials(self) -> Dict[str, str]:
...
 
     def retry(
-        self,
-        call_api: Callable[[], requests.Response],
+        self, call_method: Callable[..., requests.Response], url: str, **kwargs
     ) -> requests.Response:
         """Retry a call to an API if we get a non-ok status code.
         This function automatically retries a request if it catches a response
-        with an error code in `error_codes`. The amount of timeout also increases
+        with an error code in `error_codes`. The time interval also increases
```
bdura, line 81: The exponential backoff is not actually handled by this function, I propose we move this explanation elsewhere (see other comments)

spacy_llm/backends/rest/backend/base.py
```diff
@@ -69,31 +74,46 @@ def credentials(self) -> Dict[str, str]:
...
-        timeout = self._timeout
-        response = call_api()
+
+        def _call_api() -> requests.Response:
+            """Calls API with given timeout.
+            RETURNS (requests.Response): Response object.
+            """
+            try:
+                return call_method(url, **kwargs)
+            except (ConnectTimeout, ReadTimeout, TimeoutError) as err:
```
bdura, line 96: I'm confused, do we catch anything? These are immediately re-raised...

spacy_llm/backends/rest/backend/openai.py
```diff
@@ -87,6 +88,13 @@ def credentials(self) -> Dict[str, str]:
                 "(See OpenAI API documentation: https://platform.openai.com/docs/models/gpt-3)"
```
bdura, line 88: I'm not sure that's the intended link

spacy_llm/backends/rest/backend/openai.py
```diff
@@ -87,6 +88,13 @@ def credentials(self) -> Dict[str, str]:
...
             )
 
+        # Ensure endpoint is supported.
+        url = self._url if self._url else self.supported_models[self._config["model"]]
+        if url not in (Endpoints.non_chat, Endpoints.chat):
+            raise ValueError(
+                f"Endpoint {url} isn't supported. Please use of: {Endpoints.chat}, {Endpoints.non_chat}."
```
bdura, line 95:
```suggestion
                f"Endpoint {url} isn't supported. Please use one of: {Endpoints.chat}, {Endpoints.non_chat}."
```

spacy_llm/backends/rest/backend/openai.py
```diff
@@ -58,10 +59,10 @@ def credentials(self) -> Dict[str, str]:
...
-            lambda: requests.get(
-                "https://api.openai.com/v1/models",
-                headers=headers,
-            ),
+            call_method=requests.get,
+            url="https://api.openai.com/v1/models",
+            headers=headers,
+            timeout=self._max_request_time,
         )
         if r.status_code == 422:
```
bdura, line 67: Could we add an enum for these error codes? See [here](https://github.com/explosion/spacy-llm/pull/110/files#diff-1c49ce9ad523ef937d590bdf0fc1f62143ef7b0fea500f28d34c23d190d12e38L48).

spacy_llm/backends/rest/registry.py
```diff
@@ -11,8 +11,9 @@ def backend_rest(
     api: str,
     config: Dict[Any, Any] = SimpleFrozenDict(),
     strict: bool = True,
-    max_tries: int = 3,
-    timeout: int = 30,
+    max_tries: int = 5,
+    interval: int = 1,
```
bdura, line 15: Shouldn't `interval` be a `float`?

spacy_llm/backends/rest/backend/base.py
rmitsch: True, done in https://github.com/explosion/spacy-llm/pull/110/commits/0e8a3ca68d22ad96f9e1be5a8873d72c07cd7a9e.

spacy_llm/backends/rest/backend/base.py
rmitsch: Rephrased the error message in https://github.com/explosion/spacy-llm/pull/110/commits/0e8a3ca68d22ad96f9e1be5a8873d72c07cd7a9e. Better?

spacy_llm/backends/rest/backend/base.py
rmitsch: Updated in https://github.com/explosion/spacy-llm/pull/110/commits/f1da1268e94308420b2d90e72540ccdf02fc39cb. Thinc schedules are probably overkill if we just need this one schedule (if that changes, it'd make sense).

spacy_llm/backends/rest/backend/base.py
rmitsch: Hm, `retry()` does the [exponential backoff though](https://github.com/explosion/spacy-llm/pull/110/files#diff-1c49ce9ad523ef937d590bdf0fc1f62143ef7b0fea500f28d34c23d190d12e38R113)?

spacy_llm/backends/rest/backend/base.py
rmitsch: Here we just add a more concise error message on top of the stack trace for easier error identification.

spacy_llm/backends/rest/backend/openai.py
rmitsch: Fixed in https://github.com/explosion/spacy-llm/pull/110/commits/3b234885621900b26f543c29920e522d8892d9b1.

spacy_llm/backends/rest/backend/openai.py
rmitsch: Done in https://github.com/explosion/spacy-llm/pull/110/commits/d28a5da6eb9c4796b90136b3bc35b68f5b70d115.

spacy_llm/backends/rest/backend/base.py
bdura: Looks great!

spacy_llm/backends/rest/backend/base.py
bdura: Ah sorry about that. I had overlooked collapsed lines, perhaps?

spacy_llm/backends/rest/backend/base.py
bdura: Ah, right. I thought we'd want to retry after a timeout.

bdura (commented):
Looks good to me! The only question I have (as discussed on Slack) is about adding retries on timeout, which I think would make sense.

spacy_llm/backends/rest/backend/base.py
rmitsch: Updated in https://github.com/explosion/spacy-llm/pull/110/commits/d97c3c6e076b727a3ca7cf0c4b11c3b64c20cac6. Feel free to resolve this if you think this is sufficient.

bdura: Looks great!

rmitsch:
@bdura Any other comments or is this ready to be merged?

PR #114, opened 2023-05-23
feat: REL task

bdura:
## Description

This PR adds a new REL task.

### Types of change

This is a feature.

## Checklist

  - [x] `tests`
  - [x] `usage_example/tests`


usage_examples/rel_openai/README.md
```diff
@@ -0,0 +1,38 @@
+# Relation extraction using LLMs
+
+This example shows how you can use a model from OpenAI for relation extraction in
+zero- or few-shot settings.
```
kadarakos, line 4:
```suggestion
zero- and few-shot settings.
```

usage_examples/rel_openai/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+Here, we use the pretrained [`en_core_web_sm` model](https://spacy.io/models/en#en_core_web_sm)
```
kadarakos, line 6: At this point I think we are not recommending using the `en_core_web_sm` for anyone, the `en_core_web_lg` I think should be efficient enough for most people at this point?

adrianeboyd: I would strongly recommend using `en_core_web_md` instead of `en_core_web_lg` for things like this. Way smaller, performance nearly the same.

bdura: I changed the config to `en_core_web_md`

README.md
```diff
@@ -364,6 +364,45 @@ labels = COMPLIMENT,INSULT
...
+To perform few-shot learning, you can write down a few examples in a separate file, and provide these to be injected into the prompt to the LLM.
+The default reader `spacy.FewShotReader.v1` supports `.yml`, `.yaml`, `.json` and `.jsonl`.
+
+```json
+{"text": "Laura bought a house in Boston with her husband Mark.", "ents": [{"start_char": 0, "end_char": 5, "label": "PERSON"}, {"start_char": 24, "end_char": 30, "label": "GPE"}, {"start_char": 48, "end_char": 52, "label": "PERSON"}], "relations": [{"dep": 0, "dest": 1, "relation": "LivesIn"}, {"dep": 2, "dest": 1, "relation": "LivesIn"}]}
+{"text": "Michael travelled through South America by bike.", "ents": [{"start_char": 0, "end_char": 7, "label": "PERSON"}, {"start_char": 26, "end_char": 39, "label": "LOC"}], "relations": [{"dep": 0, "dest": 1, "relation": "Visits"}]}
+```
+
+Note: you'll need to add a NER-capable component to your spaCy pipeline
+and put it _before_ the REL component.
```
rmitsch, line 395: Let's specify a bit more what "NER-capable" means - i. e. that recognized entity spans are stored in `.ents` and that "NER-capable" could be e. g. `ner` or `llm` with the NER task.

rmitsch (commented):
Very nice work!

rmitsch:
Ah, and first we need to add `spacy download en_core_web_md` in the CI pipeline.

README.md
bdura: Thanks for the heads up. Is that better?

rmitsch:
@bdura [This](https://github.com/explosion/spacy-llm/actions/runs/5065991860/jobs/9095218237#step:5:195) is the warning I mean. Can you reproduce this locally?

spacy_llm/tasks/rel.py
```diff
@@ -0,0 +1,114 @@
...
+        self,
+        labels: str,
+        label_definitions: Optional[Dict[str, str]] = None,
+        examples: Optional[Callable[[], Iterable[Dict]]] = None,
+        normalizer: Optional[Callable[[str], str]] = None,
+        verbose: bool = False,
+    ):
+
+        if not Doc.has_extension("rel"):
+            Doc.set_extension("rel", default=[])
```
adrianeboyd, line 70: This would need to be set in `__call__` and `pipe` instead for the pipeline component.

usage_examples/rel_openai/fewshot.cfg
```diff
@@ -0,0 +1,26 @@
...
+examples = null
+
+[nlp]
+lang = "en"
+pipeline = ["ner", "llm_rel"]
+
+[components]
+
+[components.ner]
+source = "en_core_web_md"
```
adrianeboyd, line 11: This `source` does not include the vectors. If you use `spacy assemble` or `initialize` you'll get a warning, but the way you're running things with these `util` methods (that aren't really meant to be used like this) things aren't going through the normal spacy workflow+checks.

spacy_llm/tasks/rel.py
bdura: Ah right, it wont get executed on deserialization?

adrianeboyd: The problem isn't deserialization, it's multiprocessing.

bdura: Thanks. Done.

usage_examples/rel_openai/fewshot.cfg
bdura: > @bdura [This](https://github.com/explosion/spacy-llm/actions/runs/5065991860/jobs/9095218237#step:5:195) is the warning I mean. Can you reproduce this locally?

@rmitsch I imagine this is linked?

@adrianeboyd what's the proper way to use another pipeline's component through the configuration?

rmitsch: > I imagine this is linked?

I'd think so.

bdura:
Ok so for now let's leave it as it is? It requires quite an important change to use `spacy assemble` (or a custom utility), and I think it's better handled with all examples at once, in another PR?

adrianeboyd:
Add `[initialize.vectors]` the configs that source from `en_core_web_md` and then add `nlp.initialize()` to the scripts that run the examples.

adrianeboyd:
Ah, wait, no that won't quite work, either. Let me just write a util method...

rmitsch:
> Ok so for now let's leave it as it is? It requires quite an important change to use spacy assemble (or a custom utility), and I think it's better handled with all examples at once, in another PR?

Yes, we'll do that in another PR. Last todo is to [what Adriane said](https://github.com/explosion/spacy-llm/pull/114#issuecomment-1561191092), then we're good to merge.

usage_examples/rel_openai/fewshot.cfg
```diff
@@ -0,0 +1,29 @@
+[paths]
+examples = null
+
+[initialize]
+vectors = "en_core_web_md"
```
adrianeboyd, line 5: Can you move all the `[initialize]` blocks to the ends of the configs? (This is the standard config block order and I'm worried we might get confused when maintaining this version in the future.)

spacy_llm/tests/tasks/test_rel.py
```diff
@@ -0,0 +1,104 @@
...
+@pytest.mark.external
+@pytest.mark.skipif(has_openai_key is False, reason="OpenAI API key not available")
+@pytest.mark.parametrize("cfg_string", ["zeroshot_cfg_string", "fewshot_cfg_string"])
+def test_rel_predict(task, cfg_string, request):
+    """Use OpenAI to get REL results.
+    Note that this test may fail randomly, as the LLM's output is unguaranteed to be consistent/predictable
+    """
+    cfg_string = request.getfixturevalue(cfg_string)
+    orig_config = Config().from_str(cfg_string)
+    nlp = spacy.util.load_model_from_config(orig_config, auto_fill=True)
```
adrianeboyd, line 98: I suspect this isn't going to work well without the vectors? I thought about doing this initially but wasn't sure there was a good use case, but it might make sense to extend the `assemble` method to accept a `Config` in addition to a config path?

bdura: Ah you're right, I changed the example but not the test... Good catch.

bdura: I'll add an option to supply a config.

bdura: What do you think of this `config_or_path` parameter?

adrianeboyd: I'm not entirely sure that there aren't some potentially weird side effects related to interpolation and a provided config, but it seems okay to use for here for now.

adrianeboyd: I think the method gets a little weird in terms of the `str` option because it's not clear whether it's a config string or a path string.

adrianeboyd: Yeah, this is why I didn't do `Config` just now, because it was getting too messy. You can go ahead with this for now and I'll think a bit about how to make it more like spacy (probably with two separate methods).

README.md
```diff
@@ -364,6 +364,46 @@ labels = COMPLIMENT,INSULT
...
+```ini
+[components.llm.task]
+@llm_tasks = "spacy.REL.v1"
+labels = LivesIn,Visits
+```
+
+| Argument            | Type                                    | Default | Description                                                                                                          |
+| ------------------- | --------------------------------------- | ------- | -------------------------------------------------------------------------------------------------------------------- |
+| `labels`            | `str`                                   |         | Comma-separated list of relation labels.                                                                             |
+| `label_description` | `Optional[Dict[str, str]]`              | `None`  | Dictionary providind a description for each relation label.                                                          |
```
adrianeboyd, line 381:
```suggestion
| `label_description` | `Optional[Dict[str, str]]`              | `None`  | Dictionary providing a description for each relation label.                                                          |
```

rmitsch (approved):

rmitsch:
@bdura Before you merge, run the external tests again? (Locally or GH, doesn't matter)

bdura:
> @bdura Before you merge, run the external tests again? (Locally or GH, doesn't matter)

Still some failures. Thanks, I'll investigate.

PR #117, opened 2023-05-23
Fix multiprocessing with `nlp.pipe`

kabirkhan:
## Description

Currently, calling `nlp.pipe(..., n_process=2)` fails due to the way we have local functions inside our registries.
I think all that's needed is to move these private implementations into the outer scope.

Starting with failing tests for now.

## Checklist

  - [x] `tests`
  - [x] `usage_example/tests`


spacy_llm/backends/dolly.py
```diff
@@ -79,8 +79,13 @@ def backend_dolly_hf(
         config = _compile_default_config()
 
     llm_pipeline = transformers.pipeline(model=model, **config)
+    return DollyBackend(llm_pipeline).query
```
kabirkhan, line 82: If we use `functools.partial` or a `lambda` function here instead, then the `_validate_types` function can't work. We need a class and a method here to track state and have valid type annotations.

spacy_llm/backends/dolly.py
```diff
@@ -79,8 +79,13 @@ def backend_dolly_hf(
...
 
-    def query(prompts: Iterable[str]) -> Iterable[str]:
-        return [llm_pipeline(pr)[0]["generated_text"] for pr in prompts]
 
-    return query
+class DollyBackend:
```
kabirkhan, line 85: Should we have a base backend to inherit from? Should `query` here just be `__call__`?

spacy_llm/pipeline/llm.py
```diff
@@ -91,7 +91,7 @@ def _validate_types(task: LLMTask, backend: PromptExecutor) -> None:
     type_hints = {
         "template": typing.get_type_hints(task.generate_prompts),
         "parse": typing.get_type_hints(task.parse_responses),
-        "backend": typing.get_type_hints(backend),
+        "backend": typing.get_type_hints(backend.__call__) if hasattr(backend, "__call__") else typing.get_type_hints(backend),
```
kabirkhan, line 94: This allows us to pass in a class instance with a `__call__` method and get type hints for that specifically

spacy_llm/backends/dolly.py
kabirkhan: Also will add docstrings and tests if we agree on this approach

rmitsch (commented):
Thanks for catching this!

Overall I agree with your solution and don't see another way to handle this. I wasn't happy with having one additional backend per integration (MiniChain, LangChain etc.), so I abstracted this out into one separate, minimal, generic integration `Backend` class. (I'm not happy with having another layer of indirection either way, but there doesn't seem to be a way around if we want both multiprocessing and type checks to work).

This required expanding the type checks a little bit. While I was at it, I did some minor refactoring to have a cleaner separation between native (REST) backend and integrated backends.

spacy_llm/ty.py
```diff
@@ -20,3 +23,135 @@ def parse_responses(
...
+        # checking more complicated (and there probably isn't much demand for this anyway).
+        if not isinstance(backend, integration.Backend):
+            raise ValueError(
+                "Generics in backend objects are currently not supported when not using spacy_llm."
+                "backends.integration.Backend."
+            )
+        return {
+            "input": Iterable[backend.__orig_class__.__args__[1]],  # type: ignore[name-defined]
+            "return": Iterable[backend.__orig_class__.__args__[2]],  # type: ignore[name-defined]
+        }
```
rmitsch, line 73: This is new and should be reviewed. The rest of the type checking routines in `ty.py` were just moved.

spacy_llm/ty.py
```diff
@@ -20,3 +23,138 @@ def parse_responses(
...
```
rmitsch, line 76: This is new and should be reviewed. The rest of the type checking routines in `ty.py` were just moved.

spacy_llm/backends/integration/base.py
```diff
@@ -0,0 +1,22 @@
+import dataclasses
+from typing import Iterable, Generic, TypeVar, Callable
+
+# Object/Callable from third-party lib used to execute prompts.
+_Integration = TypeVar("_Integration")
+# Type of prompts returned from Task.generate_prompts().
+_PromptType = TypeVar("_PromptType")
+# Type of responses returned from query function.
+_ResponseType = TypeVar("_ResponseType")
```
kabirkhan, line 9: I don't really agree with `_PromptType` and `_ResponseType` being typevars. Like these are strs can't they just be strs?

spacy_llm/backends/rest/backend/openai.py
```diff
@@ -14,7 +14,7 @@ class Endpoints(str, Enum):
     NON_CHAT = "https://api.openai.com/v1/completions"
 
 
-class OpenAIBackend(Backend):
+class OpenAIRESTBackend(Backend):
```
kabirkhan, line 17: nit but I think we need a rename of the base rest backend as well right?

To like `RESTBackend` or something. `REST` is kind of a weird name though btw. `HttpBackend` and `OpenAIHttpBackend` might make more sense?

spacy_llm/backends/integration/dolly.py
```diff
@@ -78,9 +90,6 @@ def backend_dolly_hf(
     if not config or len(config) == 0:
         config = _compile_default_config()
 
-    llm_pipeline = transformers.pipeline(model=model, **config)
-
-    def query(prompts: Iterable[str]) -> Iterable[str]:
-        return [llm_pipeline(pr)[0]["generated_text"] for pr in prompts]
-
-    return query
+    return Backend["transformers.Pipeline", str, str](
```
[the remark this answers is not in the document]
rmitsch: Fair point regarding `transformers.Pipeline`. It's not necessary to maintain to this information for the purpose of type-checking.

We don't want to enforce `_PromptType` and `_ResponseType` to be `str` though, they can't be anything.

> Then we make a subclass for each backend to handle its init.

This would be IMO unnecessary overhead, each of these classes will do exactly the same thing.

spacy_llm/backends/rest/backend/openai.py
rmitsch: Eh, that was a refactoring slip-up. I renamed it back. I'm open to renaming REST -> HTTP though.

spacy_llm/backends/integration/dolly.py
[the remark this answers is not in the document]
rmitsch: Looking at how to do this without generics.

spacy_llm/backends/integration/base.py
bdura: I'm not an expert of other backends, but we might want to support more exotic types in the future, right? For instance including richer agent logic within the prompt might be a legitimate use case?

spacy_llm/backends/integration/dolly.py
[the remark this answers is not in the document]
bdura: As a middle ground, we could define a:

```python
from .base import Backend
from ...compat import transformers

DollyBackend = Backend["transformers.Pipeline", str, str]
```

Then the logic wouldn't be so cluttered. Although between that and a `Protocol`, the debate seems fair.

spacy_llm/backends/integration/langchain.py
```diff
@@ -55,14 +56,10 @@ def backend_langchain(
...
     if api in type_to_cls_dict:
-        backend = type_to_cls_dict[api](**config)
-
-        query_fn = query_langchain() if query is None else query
-
-        def _query(prompts: Iterable[Any]) -> Iterable[Any]:
-            return query_fn(backend, prompts)
-
-        return _query
+        return Backend["BaseLLM", Any, Any](
```
bdura, line 59: Ah, if we end up using this often then maybe it defeats the purpose of `TypeVar`s?

spacy_llm/backends/rest/__init__.py
```diff
@@ -1,3 +1,4 @@
 from .registry import backend_rest
+from .backend.base import Backend
 
-__all__ = ["backend_rest"]
+__all__ = ["backend_rest", "Backend"]
```
bdura, line 4: Why do we need to re-export `Backend`? By the way, should we rename it, eg to `RESTBackend`?

bdura (commented):
I left a few minor comments. Otherwise looks great.

spacy_llm/backends/integration/dolly.py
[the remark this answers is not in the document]
rmitsch: The core issues here are:
1. we need to use global `Callable`s to avoid the multiprocessing issue
2. we need type information to be available for our introspective type checks

`Protocol` wouldn't help us with #2. We'd need to write a separate `Protocol` per backend, which amounts to Kabir's original suggestion.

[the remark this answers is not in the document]
rmitsch: I rewrote this to work without `Generic`s. There's only a little bit of weird stuff happening in the type checks, which I think is an ok trade-off to have. 

Let me know what you think @bdura @kabirkhan.

spacy_llm/backends/integration/base.py
rmitsch: Yes, they can be any type.

spacy_llm/backends/integration/langchain.py
rmitsch: Outdated, see https://github.com/explosion/spacy-llm/pull/117#discussion_r1205157647.

spacy_llm/backends/rest/__init__.py
rmitsch: This was only done for more convenient access with the `Generic` approach. Reversed in https://github.com/explosion/spacy-llm/pull/117/commits/61f5984861e954cd570e87d47a3f92eef18ca276.

spacy_llm/backends/integration/dolly.py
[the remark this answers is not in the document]
bdura: `# type: ignore[arg-type]` is too bad...

bdura (approved):

spacy_llm/backends/integration/dolly.py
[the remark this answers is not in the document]
rmitsch: Yeah, I don't know why `mypy` complains about a type mismatch otherwise - it should recognize that these are `TypeVar`s, I think.

spacy_llm/backends/integration/base.py
kabirkhan: Ahh ok thanks for the background @bdura, makes sense

kabirkhan:
Can't approve since I opened this but looks good to me. Just left 1 comment, not sure we need a dataclass for the Backend. Also, I'm generally in favor of generics but I think this is definitely cleaner without the 3 generic params.

spacy_llm/backends/integration/base.py
```diff
@@ -0,0 +1,20 @@
+import dataclasses
+from typing import Iterable, TypeVar, Callable, Any
+
+# Type of prompts returned from Task.generate_prompts().
+_PromptType = TypeVar("_PromptType")
+# Type of responses returned from query function.
+_ResponseType = TypeVar("_ResponseType")
+
+
+@dataclasses.dataclass
```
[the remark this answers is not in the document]
rmitsch: It doesn't, I just didn't want to write `__init__()` :upside_down_face: Changed to normal class in https://github.com/explosion/spacy-llm/pull/117/commits/56f0cdfccec941f3f7508abb5b7013b4f12609fd.

rmitsch (approved):

PR #119, opened 2023-05-24
build: extra require for `langchain`, `minichain`, `transformers`

bdura:
## Description

This PR proposes three extras:
- `langchain`
- `minichain`
- `hf`, for running Hugging Face models

### Types of change

This is a change in the dependencies.

## Checklist

  - [x] `tests`
  - [x] `usage_example/tests`

### DIFF
--- a/README.md
+++ b/README.md
@@ -518,7 +518,7 @@ This allows you to have the setting `device=cuda:0` in your config, which ensure
 python -m pip install "torch>=1.13.1,<2.0"
 python -m pip install "transformers>=4.28.1,<5.0"
 # Or install with spacy-llm directly
-python -m pip install "spacy-llm[hf]"
+python -m pip install "spacy-llm[transformers]"
 ```
 
 If you don't have access to a GPU, you can install `accelerate` and set`device_map=auto` instead, but be aware that this may result in some layers getting distributed to the CPU or even the hard drive,
--- a/README.md
+++ b/README.md
@@ -454,6 +454,8 @@ To use [MiniChain](https://github.com/srush/MiniChain) for the API retrieval par
 
 ```shell
 python -m pip install "minichain>=0.3,<0.4"
+# Or install with spacy-llm directly
+python -m pip install "spacy-llm[minichain]"
 ```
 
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -39,6 +39,16 @@ install_requires =
...
 
+[options.extras_require]
+minichain =
+    minichain>=0.3,<0.4
+langchain =
+    langchain>=0.0.144,<0.1
+hf =
+    cupy-cuda11x
+    torch>=1.13.1,<2.0
+    transformers>=4.28.1,<5.0
```
adrianeboyd, line 50: To be parallel to the spacy extras, I would call this option `transformers` and leave out `cupy`. (I don't think you need it and if you do, `cupy-cuda11x` is too specific as a version. I can make suggestions if you do need to keep it here.)

bdura: Good point. I aligned the packages on the README, maybe there's something to change there as well.

README.md
```diff
@@ -514,6 +518,8 @@ This allows you to have the setting `device=cuda:0` in your config, which ensure
 python -m pip install "cupy-cuda11x"
 python -m pip install "torch>=1.13.1,<2.0"
 python -m pip install "transformers>=4.28.1,<5.0"
+# Or install with spacy-llm directly
+python -m pip install "spacy-llm[hf]"
```
adrianeboyd, line 522:
```suggestion
python -m pip install "spacy-llm[transformers]"
```

adrianeboyd:
You need cupy for thinc on GPU, but for transformers it only depends on your torch install.

rmitsch (commented):
Looks good!

> You need cupy for thinc on GPU, but for transformers it only depends on your torch install.

@adrianeboyd How should we reflect this in `setup.cfg`?

adrianeboyd:
For `spacy-llm` I don't think you need to worry about cupy in general?

Very probably people will want more control over the torch devices, which we'd usually do through `require_gpu` and `require_cpu` with spacy in general, but I would not worry about this for this PR.

PR #120, opened 2023-05-24
Make cache object configurable

rmitsch:
## Description

Make cache object configurable by accepting a resolved cache object instead of a dict. Adds a `Protocol` cache objects have to adhere to.

### Types of change

Chore.

## Checklist

  - [x] `tests`
  - [x] `usage_example/tests`


spacy_llm/cache.py
```diff
@@ -5,8 +5,21 @@
...
+    path: Optional[Union[str, Path]],
+    batch_size: int,
+    max_batches_in_mem: int,
+):
+    return BatchCache(
+        path=path, batch_size=batch_size, max_batches_in_mem=max_batches_in_mem
+    )
+
+
+class BatchCache:
```
bdura, line 22: Is there a particular reason that we separate the factory from the class itself?

```suggestion
@registry.llm_misc("spacy.BatchCache.v1")
class BatchCache:
```

bdura (approved):

spacy_llm/cache.py
rmitsch: > ...a policy that every factory should be a function and not a class could make sense for consistency's sake.

Yes, exactly that. There's no particular reason to do this here, it's to be consistent with the upcoming changes in https://github.com/explosion/spacy-llm/pull/81.

spacy_llm/pipeline/llm.py
```diff
@@ -196,19 +199,13 @@ def __init__(
...
         self._task = task
         self._backend = backend
-        self._cache = Cache(
-            path=cache["path"],
-            batch_size=int(cache["batch_size"]),
-            max_batches_in_mem=int(cache["max_batches_in_mem"]),
-            vocab=vocab,
-        )
+        self._cache = cache
+        self._cache.vocab = vocab
```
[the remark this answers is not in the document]
rmitsch: Hm, I see what you mean. I'll go ahead with this, but I'm open to changing this later.

PR #121, opened 2023-05-24
Add assemble util method

adrianeboyd:
## Description

This is a non-CLI version of `spacy assemble` that returns the pipeline object rather than saving it to a directory.

### Types of change

Enhancement.

## Checklist

  - [ ] `tests`
  - [ ] `usage_example/tests`

### DIFF
--- a/spacy_llm/util.py
+++ b/spacy_llm/util.py
@@ -1,0 +1,21 @@
+from typing import Any, Dict, Union
+from pathlib import Path
+
+from spacy.util import SimpleFrozenDict, get_sourced_components, load_config
+from spacy.util import load_model_from_config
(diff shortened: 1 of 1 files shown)


bdura (approved):
Thanks a lot!

adrianeboyd:
It probably makes sense to refactor `spacy.cli.assemble.assemble_cli` so that this is a spacy util method rather than a local util method, but I think it would make current development easier to go ahead and include it here until it's available in a better form from spacy.

adrianeboyd:
(And in spacy it's kind of entangled with some CLI settings, so it's a little tricky to separate cleanly.)

PR #124, opened 2023-05-25
Use uint64 range for cache hash keys

adrianeboyd:
## Description

Use uint64 range for cache hash keys.

### Types of change

Enhancement.

## Checklist

  - [x] `tests`
  - [ ] `usage_example/tests`


bdura (approved):

adrianeboyd:
I don't really know about the tests around caching, but I think this is basically what you originally meant to do with this `sum` as a hash key?

PR #127, opened 2023-05-25
feat: add optional prompt storage support

bdura:
## Description

This PR adds prompt storage support. Prompt and responses are optionally saved to `doc._.llm_io`

### Types of change

This is a feature.

## Checklist

  - [x] `tests`
  - [x] `usage_example/tests`


spacy_llm/backends/rest/backend/noop.py
```diff
@@ -3,6 +3,8 @@
 
 from .base import Backend
 
+NOOP_RESPONSE = ""
```
rmitsch, line 6: Let's make this a static class variable, if there's no particular reason to have it global.

spacy_llm/pipeline/llm.py
```diff
@@ -232,15 +237,22 @@ def pipe(self, stream: Iterable[Doc], *, batch_size: int = 128) -> Iterator[Doc]
...
                 doc for doc, cached_doc in zip(doc_batch, is_cached) if not cached_doc
             ]
             try:
-                prompts = self._task.generate_prompts(noncached_doc_batch)
-                responses = self._backend(prompts)
+                prompts = list(self._task.generate_prompts(noncached_doc_batch))
+                iter_prompts = iter(prompts)
+
+                responses = list(self._backend(prompts))
+                iter_responses = iter(responses)
```
rmitsch, line 254: Why the list conversion?

spacy_llm/tasks/noop.py
```diff
@@ -4,12 +4,14 @@
 
 from ..registry import registry
 
+NOOP_PROMPT = "Don't do anything."
```
rmitsch, line 7: Variables not needed outside of the module (ignoring tests) should be prefixed with `_`.

rmitsch (commented):
Looks good. Docs should be updated too.

spacy_llm/pipeline/llm.py
bdura: Otherwise the iterator is consumed by `self._backend(prompts)`. Unless using `iter` makes it unnecessary?

spacy_llm/tasks/noop.py
bdura: Done.

spacy_llm/pipeline/llm.py
adrianeboyd: I think what you're looking for might be `itertools.tee`?

rmitsch: Ah, you're using the prompts and responses for the logging again. Give it a shot with `itertools.tee()`?

bdura: Since they are consumed sequentially [I'm not sure it will help](https://docs.python.org/3/library/itertools.html#itertools.tee:~:text=This%20itertool%20may%20require%20significant%20auxiliary%20storage%20(depending%20on%20how%20much%20temporary%20data%20needs%20to%20be%20stored).%20In%20general%2C%20if%20one%20iterator%20uses%20most%20or%20all%20of%20the%20data%20before%20another%20iterator%20starts%2C%20it%20is%20faster%20to%20use%20list()%20instead%20of%20tee().)...

adrianeboyd: It's not that this doesn't work, but I think it's more confusing at first glance than using `itertools.tee`.

I do worry in general that this whole collection of multiple interleaved steps with iterators is very brittle and likely to get out-of-sync.

adrianeboyd: It might also be clearer if the second iterators are only created if required?

bdura: I agree. We could also defer the responsibility to the task... I did it this way because it does not require any change on the task, but come to think of it it would make sense that they handle it.

Consider the case where the `Prompt` is a complex, non-serialisable object... It should be handled differently.

bdura: Ah, and I also struggled to include it at the task level _and_ get the component name...

rmitsch: > Consider the case where the Prompt is a complex, non-serialisable object... It should be handled differently.

We could only store the stringified version of prompts and responses?

rmitsch: > Ah, and I also struggled to include it at the task level and get the component name...

Can you elaborate on this?

bdura: > We could only store the stringified version of prompts and responses?

Good idea. Then it's the task's/backends responsibility to generate prompts/responses that display a proper `__str__` method.

bdura: I looked at a potential solution to implement prompt/response storage at the task level, but it looks very convoluted. We'd likely need to apply substantial changes to the architecture.

> Ah, and I also struggled to include it at the task level and get the component name...

Storage is done on a per-component basis so we need the component name. We could use a list or more exotic workarounds, but I think that's more confusing than anything else. Do you have another idea?

spacy_llm/pipeline/llm.py
```diff
@@ -126,25 +134,42 @@ def pipe(self, stream: Iterable[Doc], *, batch_size: int = 128) -> Iterator[Doc]
...
             noncached_doc_batch = [
                 doc for doc, cached_doc in zip(doc_batch, is_cached) if not cached_doc
             ]
             try:
-                prompts = self._task.generate_prompts(noncached_doc_batch)
-                responses = self._backend(prompts)
+                # We need to use tee to store generated prompts/responses
+                # Otherwise, they will be consumed
+                prompts = tee(self._task.generate_prompts(noncached_doc_batch))
+                responses = tee(self._backend(prompts))
```
adrianeboyd, line 150: Not with this naming, but what I meant was more like:

```suggestion
                prompts = tee(self._task.generate_prompts(noncached_doc_batch))
                if self._save_io:
                    prompts, prompts_for_saving = tee(prompts)
                responses = self._backend(prompts))
                if self._save_io:
                    responses, responses_for_saving = tee(responses)
```

spacy_llm/pipeline/llm.py
```diff
@@ -135,16 +146,32 @@ def pipe(self, stream: Iterable[Doc], *, batch_size: int = 128) -> Iterator[Doc]
             try:
                 prompts = self._task.generate_prompts(noncached_doc_batch)
                 responses = self._backend(prompts)
+
+                if self._save_io:
+                    prompts, iter_prompts = tee(prompts)
```
adrianeboyd, line 151: `prompts` is already consumed at this point?

spacy_llm/pipeline/llm.py
```diff
@@ -135,16 +146,32 @@ def pipe(self, stream: Iterable[Doc], *, batch_size: int = 128) -> Iterator[Doc]
...
+                    responses, iter_responses = tee(responses)
+
                 modified_docs = iter(
                     self._task.parse_responses(noncached_doc_batch, responses)
                 )
+
+                if self._save_io:
+                    iter_prompts = iter(prompts)
+                    iter_responses = iter(responses)
+
```
adrianeboyd, line 162: And then you don't need this.

bdura: Right, I saw that... I pushed the necessary commits.

spacy_llm/pipeline/llm.py
```diff
@@ -134,17 +145,30 @@ def pipe(self, stream: Iterable[Doc], *, batch_size: int = 128) -> Iterator[Doc]
             ]
             try:
                 prompts = self._task.generate_prompts(noncached_doc_batch)
+                if self._save_io:
+                    prompts, iter_prompts = tee(prompts)
```
adrianeboyd, line 149: This is very confusing naming.

bdura: Agreed. `saved_prompts` ?

README.md
```diff
@@ -213,6 +213,12 @@ Each `llm` component is defined by two main settings:
 Moreover, the component also implements [**caching**](#cache) functionality to avoid running
 the same document through an LLM service (be it local or through a REST API) more than once.
 
+Finally, you can choose to save a stringified version of LLM prompts/responses
```
adrianeboyd, line 216: I think the README could use a table of `llm` config settings?

bdura: I agree, it's overdue. I added a table and kept the two main settings apart like before, it seems important. Let me know what you think.

README.md
```diff
@@ -203,16 +203,32 @@ my_other_config_val = 0.3
 
 ## 📓 API
 
-Each `llm` component is defined by two main settings:
+`spacy-llm` exposes a `llm` factory that accepts the following configuration options:
+
+| Argument  | Type                                        | Description                                                                         |
+| --------- | ------------------------------------------- | ----------------------------------------------------------------------------------- |
+| `name`    | `str`                                       | The name of the component.                                                          |
```
adrianeboyd, line 210: Ah, `name` shouldn't be one of the config settings. Let me check this in the refactoring...

adrianeboyd: It's fine, just `nlp` and `name` aren't in the config as config settings.

```suggestion
```

bdura: Thanks !

spacy_llm/pipeline/llm.py
```diff
@@ -126,6 +134,9 @@ def pipe(self, stream: Iterable[Doc], *, batch_size: int = 128) -> Iterator[Doc]
         batch_size (int): The number of documents to buffer.
         YIELDS (Doc): Processed documents in order.
         """
+        if not Doc.has_extension("llm_io"):
```
kabirkhan, line 137: This needs to be added to `__call__` as well. Since our backend + response parser always work on batches I think putting this in a util method would work best. 

We're deferring this logging PR (#80) for 0.3.0 but the `_process_docs` method could be added to this PR I think. Then the extension installation (+ setting of the prompt/response) only needs to happen in one spot.

[_process_docs method](https://github.com/explosion/spacy-llm/blob/3e7c481a0300f3fd227df98b19785373305c3c81/spacy_llm/pipeline/llm.py#L242)

bdura: Ah absolutely, thanks! I'll wait for your PR to be merged then.

bdura:
Waiting for #80 to be merged.

spacy_llm/pipeline/llm.py
rmitsch: Ah, ok. I think it's fine to do this at the component level. I don't think customized logging (via tasks) is a priority right now.

spacy_llm/pipeline/llm.py
rmitsch: I agree with Kabir, but I don't think we need to wait for https://github.com/explosion/spacy-llm/pull/80 to be merged. I'd recommend copying over `_process_docs()` and adding setting the extension.

kabirkhan: Oh yeah, that's what I meant, I think this PR should be merged first but we should copy over the core of `_process_docs` so we only need to install the extensions in 1 method.

spacy_llm/pipeline/llm.py
```diff
@@ -134,17 +145,30 @@ def pipe(self, stream: Iterable[Doc], *, batch_size: int = 128) -> Iterator[Doc]
             ]
             try:
                 prompts = self._task.generate_prompts(noncached_doc_batch)
+                if self._save_io:
+                    prompts, saved_prompts = tee(prompts)
```
kabirkhan, line 149: I don't think `itertools.tee` helps here since we consume the iterator in the backend but I've never properly profiled this so I'm not sure

spacy_llm/pipeline/llm.py
```diff
@@ -37,6 +38,7 @@ class CacheConfigType(TypedDict):
             "batch_size": 64,
             "max_batches_in_mem": 4,
         },
+        "save_io": False,
```
kabirkhan, line 41: I personally think we should default this to True. Is memory consumption/storage space the only concern here?

kabirkhan:
See https://github.com/explosion/spacy-llm/pull/147 for refactored `_process_docs` method so we only need to install extension attrs there, not separately in `__call__` and `pipe`

spacy_llm/pipeline/llm.py
rmitsch: I think so. I'm on the fence about whether to make this opt-out.

spacy_llm/pipeline/llm.py
rmitsch: Performance isn't that critical here anyway, I believe.

spacy_llm/pipeline/llm.py
bdura: Alright!

adrianeboyd:
One concern is what happens if this is incorporated into spacy at some point: spacy itself has a policy of never setting any custom extensions. There are a few tokenizers that add data to `doc.user_data` (the same place extension data is stored), but not in a way that interacts with custom extensions data. The idea is that these might conflict with user-provided extensions so it's better if spacy never conflicts with any potential user code here.

rmitsch:
https://github.com/explosion/spacy-llm/pull/147 has been merged (but you already saw that).

spacy_llm/pipeline/llm.py
bdura: Done, thanks.

bdura:
> One concern is what happens...

I changed it to use `doc.user_data` directly. Let me know.

@rmitsch there's still the opt-in/out decision to make. Otherwise I think we're good to go.

rmitsch:
I'm inclined to leave it as `False`. We can also change that in a later PR, if we feel that's better.

spacy_llm/pipeline/llm.py
```diff
@@ -118,12 +126,43 @@ def pipe(self, stream: Iterable[Doc], *, batch_size: int = 128) -> Iterator[Doc]
...
+            #         else:
+            #             doc = next(modified_docs)
+
+            #             if self._save_io:
+            #                 llm_io = doc._.llm_io[self._name]
+            #                 llm_io["prompt"] = str(next(saved_prompts))
+            #                 llm_io["response"] = str(next(saved_responses))
+
+            #             self._cache.add(doc)
+            #             yield doc
```
rmitsch, line 165: Should be removed, no?

rmitsch (commented):
LGTM for me except for the commented code. I'd want @adrianeboyd to sign off on storing this in `user_data` before we merge though.

spacy_llm/pipeline/llm.py
bdura: Gosh. Thanks

bdura: Done.

spacy_llm/pipeline/llm.py
bdura: It will be opt-in for now (see [comment](https://github.com/explosion/spacy-llm/pull/127#issuecomment-1569706839))

spacy_llm/pipeline/llm.py
bdura: Exactly. See [comment](https://github.com/explosion/spacy-llm/pull/127#discussion_r1206379607).

adrianeboyd:
So the only places does this are pretty obscure, just the Japanese and Korean tokenizers, but I think this is okay.

It is kind internals, so I think it would make sense to provide utility functions (+ usage examples) that can be used to access this data with custom extension getters. So you could do something like:

```python
doc._.llm_io(pipe_name)["prompt"]
```

or

```python
doc._.llm_io(pipe_name).prompt
```

bdura:
@adrianeboyd in that case maybe using `doc._.llm_io` directly makes more sense for now (since it's conceptually simpler)? If we're going to use a custom extension in both cases...

rmitsch:
@adrianeboyd Sorry, it's not quite clear to me what you are recommending. Should we use `user_data` or `_.llm_io`?

adrianeboyd:
Let's stick with `user_data` for this PR. The idea with custom extensions would be that users would have to run some util method to add the custom extensions, just for easier access to this data, and it would be easy to add something like that later.

bdura:
I pushed a change to remove any mention of `Doc._.llm_io`. I think we're good to go, although if you have a better suggestion regarding the naming conventions I'd love your input.

bdura:
By the way @rmitsch: watch out, this is set up to merge into main. There's nothing breaking though, so we should be fine – just thought I'd mention it nonetheless.

adrianeboyd:
Since this is adding a config setting to `llm`, I think it should go in the next minor version.

adrianeboyd:
Ah, too many differences `develop` to switch just in the UI here, sorry.

rmitsch:
@bdura Can you target this PR at `develop` and resolve the conflicts?

rmitsch (approved):
Great work! Merging after the tests pass.

bdura:
Give me a minute :)

PR #128, opened 2023-05-25
ci: add caching to validation workflow

bdura:
## Description

This PR adds caching to the validation workflow for faster CI runs.

### Types of change

This is a CI change.

## Checklist

  - [ ] `tests`
  - [ ] `usage_example/tests`

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -13,11 +13,20 @@ jobs:
           python-version: 3.9
           cache: "pip" # caching pip dependencies
 
-      - name: Install dependencies
+      - name: Set env variables
         run: |
-          pip install --upgrade pip
-          pip install -r requirements.txt
-          pip install -r requirements-dev.txt
+          echo "PY=$(python -VV | sha256sum | cut -d' ' -f1)" >> $GITHUB_ENV
+          echo "PRECOMMIT=$(cat requirements-dev.txt | grep pre-commit)" >> $GITHUB_ENV
+
+      - uses: actions/cache@v3
+        with:
+          path: ~/.cache/pre-commit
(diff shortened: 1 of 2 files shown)


.github/workflows/validate.yml
```diff
@@ -13,11 +13,20 @@ jobs:
...
+
+      - name: Set env variables
+        run: |
+          echo "PY=$(python -VV | sha256sum | cut -d' ' -f1)" >> $GITHUB_ENV
+          echo "PRECOMMIT=$(pip list | grep pre-commit)" >> $GITHUB_ENV
+
+      - uses: actions/cache@v3
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit|${{ env.PY }}|${{ env.PRECOMMIT }}|${{ hashFiles('.pre-commit-config.yaml') }}
```
rmitsch, line 29: Can you elaborate on what's happening here?

bdura: Sure. GHA's caching mechanism relies on a string key to perform cache creation and retrieval.

If we break down the `pre-commit|${{ env.PY }}|${{ env.PRECOMMIT }}|${{ hashFiles('.pre-commit-config.yaml') }}` key, it means that the cache will be invalidated every time the Python version, pre-commit version or pre-commit config changes.

bdura: Then the cache works roughly like this:

1. Check the key, if it exists, retrieve the cache and restore it at the given path (here `~/.cache/pre-commit`, the default pre-commit cache folder)
2. Run other actions
3. At the end of the workflow, run the "Post Run Action" which creates the cache if there was no hit

rmitsch: Thanks!

rmitsch (approved):

PR #132, opened 2023-05-26
Fix type check issues

rmitsch:
## Description

Fix type check issues (for LangChain integration).

### Types of change

Bug fix.

## Checklist

  - [x] `tests`
  - [x] `usage_example/tests`


spacy_llm/backends/integration/langchain.py
```diff
@@ -21,22 +21,28 @@ def _check_installation() -> None:
 
 
 @registry.llm_queries("spacy.CallLangChain.v1")
-def query_langchain() -> Callable[["BaseLLM", Iterable[Any]], Iterable[Any]]:
+def query_langchain() -> Callable[
+    ["langchain.llms.base.BaseLLM", Iterable[Any]], Iterable[Any]
```
adrianeboyd, line 25: Don't you need to update the `TYPE_CHECKING` block to correspond to this?

rmitsch: I don't think it's necessary at all anymore now?

pyproject.toml
```diff
@@ -15,8 +15,9 @@ testpaths = ["tests"]
 filterwarnings = [
     "error",
     "ignore:^.*pkg_resources.*:DeprecationWarning",
-    "ignore::sqlalchemy.exc.MovedIn20Warning"
+    "ignore::sqlalchemy.exc.MovedIn20Warning",
+    "ignore:^.*You are trying to use a chat model. This way of initializing it is no longer supported. Instead, please use.*:UserWarning",
```
rmitsch, line 19: LangChain prefers a different way of initializing OpenAI chat models now. The old one still works however.

bdura: It seems like you did modify the instantiation method though... Should we keep the last exception?

spacy_llm/backends/integration/langchain.py
```diff
@@ -50,14 +52,20 @@ def backend_langchain(
...
+    if "model" not in config:
+        raise ValueError("The LLM model must be specified in the config.")
+
     # langchain.llms.type_to_cls_dict contains all API names in lowercase, so this allows to be more forgiving and make
     # "OpenAI" work as well "openai".
     api = api.lower()
-    type_to_cls_dict: Dict[str, Type[BaseLLM]] = langchain.llms.type_to_cls_dict
+    type_to_cls_dict: Dict[
+        str, Type[langchain.llms.base.BaseLLM]
+    ] = langchain.llms.type_to_cls_dict
```
bdura, line 63: Why is this necessary?

bdura (commented):
Looks good. I think we can remove the error-on-warning exception, and I'm not sure whether re-defining the models dict actually helps - but I trust you.

spacy_llm/backends/integration/langchain.py
rmitsch: IIRC `mypy` complained otherwise.

pyproject.toml
rmitsch: No, I'm just passing `model_name` as separate arg. What LangChain actually wants us to do is to instantiate a differently named class. The current approach does exactly the same behind the LangChain curtains, but with a warning and rerouting to this new class with a warning.

What we should actually do is to figure out a new way how to instantiate the correct `LangChain` classes generically (like it's done currently with `langchain.llms.type_to_cls_dict`. I haven't looked into this and would do so in a separate PR.

spacy_llm/backends/integration/langchain.py
bdura: Perfect then!

bdura (approved):

PR #135, opened 2023-05-26
feat: scoring

bdura:
## Description

This PR adds basic scoring capabilities.

### Types of change

This is a feature.

## Checklist

  - [x] `tests`
  - [x] `usage_example/tests`


bdura:
@rmitsch I'd love to have your input on this.

spacy_llm/tests/tasks/test_textcat.py
```diff
@@ -546,3 +553,65 @@ def test_external_template_actually_loads():
...
+        reference = predicted.copy()
+
+        reference.cats = {"Insult": 1.0 if i < n_insults else 0.0}
+
+        examples.append(Example(predicted, reference))
+
+    scores = nlp.evaluate(examples)
+
+    pos = n_insults / len(INSULTS)
+    1 - pos
```
adrianeboyd, line 614: ?

```suggestion
```

rmitsch (commented):
Looks good! Approving after Adriane's comments have been addressed.

spacy_llm/tests/tasks/test_textcat.py
bdura: Ah thanks. My linter removed the variable name because I never used it.

bdura:
Thanks for the review, the comments have been addressed – I think we can merge once tests pass.

spacy_llm/tasks/ner.py
```diff
@@ -114,3 +115,9 @@ def assign_spans(
     ) -> None:
         """Assign spans to the document."""
         doc.set_ents(filter_spans(spans))
+
+    def scorer(
+        self,
+        examples: Iterable[Example],
+    ) -> Dict[str, Any]:
+        return get_ner_prf(examples, attr="ents")
```
adrianeboyd, line 123:
```suggestion
        return get_ner_prf(examples)
```

rmitsch (approved):
Great to to have this! Needs to be synced with `main`, I'll merge after that.

spacy_llm/tasks/spancat.py
```diff
@@ -111,3 +112,13 @@ def assign_spans(
     ) -> None:
         """Assign spans to the document."""
         doc.spans[self._spans_key] = sorted(spans)  # type: ignore [type-var]
+
+    def scorer(
+        self,
+        examples: Iterable[Example],
+    ) -> Dict[str, Any]:
+        return spancat_score(
```
adrianeboyd, line 120: Why `spancat_score` instead of `Scorer.score_spans`? (It seems odd that this isn't the same as the other tasks.)

bdura: Because `spancat_score` implements the right getter out of the box. It's also the default scorer for spaCy's `spancat` component.

spacy_llm/tasks/ner.py
rmitsch: @bdura Let's accept this (if you don't see a reason not to), then I'll merge.

spacy_llm/tasks/spancat.py
adrianeboyd: But there are also default scorers for `ner` and `textcat` and `textcat_multilabel`?

adrianeboyd: Maybe we should go ahead and do this as registered functions, which would reduce the potential future confusion here, I think.

adrianeboyd: The problem is passing the settings in, which you don't really have access to in the config. Ugh.

bdura: Ah, right. I used `get_ner_prf` for the NER task as you suggested and the scorers for `textcat`/`textcat_multilabel` only differ on the `multi_label` argument, so that seemed cleaner to use...

rmitsch: Which settings do you not have access to in the config?

bdura: > Ah, right. I used get_ner_prf for the NER task as you suggested and the scorers for textcat/textcat_multilabel only differ on the multi_label argument, so that seemed cleaner to use...

Sorry I was responding to a previous comment, my PR page was outdated...

bdura: > Maybe we should go ahead and do this as registered functions, which would reduce the potential future confusion here, I think.

It would be nice to be able to customize the scoring function

rmitsch: Added an internal ticket to follow up on this.

adrianeboyd: The problem is that the scoring methods were originally written to be used from within the pipeline component itself, which could provide labels, settings, etc. If you move them to registered functions, then all those settings need to be provided to the registered function somehow. There's a terrible hack in spacy that passes all the settings in `pipe.cfg` from the component to the scorers, but that's not even possible from within the task here.

PR #139, opened 2023-05-28
Bump version to 0.2.0

rmitsch:
## Description

Bump version to 0.2.0.

### Types of change

Chore.

## Checklist

  - [ ] `tests`
  - [ ] `usage_example/tests`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [metadata]
-version = 0.1.2
+version = 0.2.0
 description = Integrating LLMs into structured NLP pipelines
 author = Explosion
 author_email = contact@explosion.ai


adrianeboyd:
I don't think that the `--installed` marker is a good idea. "Installed" should be the default for the test suite.

setup.cfg
```diff
@@ -39,8 +39,8 @@ install_requires =
 spacy_factories =
     llm = spacy_llm.pipeline.llm:make_llm
 spacy_misc =
-    spacy.FewShotReader.v1 = spacy_llm.registry:fewshot_reader
-    spacy.FileReader.v1 = spacy_llm.registry:file_reader
+    spacy.FewShotReader.v1 = spacy_llm.registry.fewshot_reader
+    spacy.FileReader.v1 = spacy_llm.registry.file_reader
```
adrianeboyd, line 43: Was this change just temporary to test the test suite?

rmitsch: Yes. Will reverse.

rmitsch:
> I don't think that the --installed marker is a good idea. "Installed" should be the default for the test suite.

Fair point. Instead I will adjust the external and GPU workflow to work with the installed package instead.

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [metadata]
-version = 0.1.2
+version = 0.2.0
```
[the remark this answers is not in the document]
rmitsch: Reverted in https://github.com/explosion/spacy-llm/pull/139/commits/81828ef8440d0b4b06fe6871b056bf0bde2d581a.

.github/workflows/external.yml
```diff
@@ -21,20 +23,47 @@ jobs:
...
-      - name: Install dependencies
+      - name: Build sdist
         run: |
-          pip install --upgrade pip
-          pip install -r requirements.txt
-          pip install -r requirements-dev.txt
+          python -m pip install --upgrade pip
+          python -m pip install -U build pip setuptools
+          python -m pip install -U -r requirements.txt
+          python -m build --sdist
```
adrianeboyd, line 31:
```suggestion
          python -m pip install -U build pip setuptools
          python -m build --sdist
```

.github/workflows/external.yml
```diff
@@ -21,20 +23,47 @@ jobs:
...
+      - name: Test import
+        shell: bash
+        run: |
+          python -c "import $MODULE_NAME" -Werror
+
+      - name: Install test requirements
+        run: |
+          python -m pip install -U -r requirements.txt
+          python -m pip install -U -r requirements-dev.txt
           python -m spacy download en_core_web_md
```
adrianeboyd, line 58: There is some chance that `spacy download` (without `--no-deps`) would clobber some of the other requirements, so I think it's better to prioritize the local requirements. It shouldn't, but it could be hard to catch here if it does.

```suggestion
          python -m spacy download en_core_web_md
          python -m pip install -U -r requirements.txt -r requirements-dev.txt
```

.github/workflows/external.yml
```diff
@@ -21,20 +23,47 @@ jobs:
...
+      - name: Delete source directory
+        shell: bash
+        run: |
+          rm -rf $MODULE_NAME
+
+      - name: Uninstall all packages
+        run: |
+          python -m pip freeze > installed.txt
+          python -m pip uninstall -y -r installed.txt
+
```
adrianeboyd, line 42: Since we're not running mypy or anything else from non-package requirements, this isn't really doing anything.

```suggestion
```

PR #146, opened 2023-05-30
Drop requirement for `sqlalchemy`

rmitsch:
## Description

Currently `sqlalchemy` is necessary for the test suite due to checking for a specific `sqlalchemy` warning. This compares for a check with an error message regex instead.

### Types of change

Fix.

## Checklist

  - [ ] `tests`
  - [ ] `usage_example/tests`


requirements-dev.txt
```diff
@@ -7,7 +7,7 @@ mypy>=0.990,<1.1.0; platform_machine != "aarch64" and python_version >= "3.7"
 black==22.3.0
 types-requests==2.28.11.16
 # Prompting libraries needed for testing
-langchain>=0.0.180,<0.1.0; python_version>="3.9"
+langchain>=0.0.180,<0.0.185; python_version>="3.9"
```
adrianeboyd, line 10: Where is this needed? (It's not in `setup.cfg`.)

rmitsch: What exactly?

rmitsch:
I tested the modified `pytest` warning filter locally with Python 3.11 - the warning is ignored and tests succeed.

PR #147, opened 2023-05-30
add util method to component for processing list of docs

kabirkhan:
## Description

Refactor duplicate logic in `__call__` and `pipe` into a single method that works on a list of docs for all processing logic.
This allows any new logic we add to the LLM component to just exist in 1 place. Similar to our `predict` -> `set_annotations` workflow in other components.

`__call__` just puts a single doc in a list and returns the first element in resulting list
`pipe` handles batching and exception handling but defers all prediction to `_process_docs`.

Unblocks #127 and #80 

### Types of change

Refactor/enhancement

## Checklist

  - [x] `tests`
  - [x] `usage_example/tests`

### DIFF
--- a/spacy_llm/cache.py
+++ b/spacy_llm/cache.py
@@ -55,7 +55,9 @@ class BatchCache:
         # Statistics.
         self._stats: Dict[str, int] = {
             "hit": 0,
+            "hit_contains": 0,
             "missed": 0,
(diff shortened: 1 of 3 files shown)


bdura (approved):

rmitsch (approved):
Good idea to factor this out!

PR #149, opened 2023-05-31
Allow support of custom REST backend classes

rmitsch:
## Description

Allow support of custom REST backend classes. As discussed with @bdura it hasn't been possible so far for users to add their own REST API backend implementations (without monkeypatching `supported_apis` at least). This modifies the REST backend registry function to accept `Type[spacy_llm.backends.rest.backend.base.Backend]` - this enables passing REST backend classes via registry functions.

### Types of change

Enhancement.

## Checklist

  - all tests requiring a GPU

### DIFF
--- a/spacy_llm/backends/rest/registry.py
+++ b/spacy_llm/backends/rest/registry.py
@@ -1,14 +1,14 @@
-from typing import Any, Callable, Dict, Iterable
+from typing import Any, Callable, Dict, Iterable, Union, Type
 
 from spacy.util import SimpleFrozenDict
 
-from .backend import supported_apis
+from .backend import supported_apis, base
 from ...registry import registry
 
 
 @registry.llm_backends("spacy.REST.v1")
 def backend_rest(
-    api: str,
+    api: Union[str, Type[base.Backend]],
     config: Dict[Any, Any] = SimpleFrozenDict(),
     strict: bool = True,
     max_tries: int = 5,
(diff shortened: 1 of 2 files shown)


spacy_llm/backends/rest/registry.py
```diff
@@ -30,7 +32,7 @@ def backend_rest(
         Backend instance.
     """
 
-    backend = supported_apis[api](
+    backend = (supported_apis[api] if isinstance(api, str) else api)(
```
svlandeg, line 35: I'd prefer making this more explicit for maintainability & readability purposes:
```suggestion
    backend_class = api
    if isinstance(api, str):
        backend_class = supported_apis[api]
    backend = backend_class(
```

svlandeg (commented):
I'm still a little conflicted around this.

I get that this is done to make `spacy.REST.v1` more flexible and extensible, but it also means that you can change the `Backend` implementation without bumping the `v1` version, thus circumventing the versioning.

In my view, the extensibilty & flexibility of `spacy-llm` comes from using the registry functions and allowing users to implement their own. Going in and changing/extending built-in functionality feels like a second, alternative way of doing it. Maybe it's more intuitive to users this way? But it still feels a bit confusing to me - this is not a pattern that we're using anywhere else AFAIK.

rmitsch:
Closing this after internally discussing design changes.

PR #152, opened 2023-05-31
feat: labels from examples

bdura:
## Description

This PR adds an `initialize` method to LLMWrapper, letting it use examples to create labels.

### Types of change

This is a new feature.

## Checklist

  - all tests requiring a GPU

### DIFF
--- a/spacy_llm/pipeline/llm.py
+++ b/spacy_llm/pipeline/llm.py
@@ -1,11 +1,23 @@
 from pathlib import Path
-from typing import Dict, Iterable, Iterator, List, Optional, Tuple, Union, cast
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Tuple,
+    Union,
+    cast,
+)
 
 import spacy
 from spacy.language import Language
 from spacy.pipeline import Pipe
 from spacy.tokens import Doc
 from spacy.training import Example
+from spacy.ty import InitializableComponent
 from spacy.vocab import Vocab
 
 from .. import registry  # noqa: F401
@@ -163,6 +175,19 @@ class LLMWrapper(Pipe):
                 final_docs.append(doc)
 
         return final_docs
+
+    def initialize(
+        self,
+        get_examples: Callable[[], Iterable["Example"]],
+        nlp: Iterable["Example"],
+        **kwargs: Any,
+    ):
+        if isinstance(self._task, InitializableComponent):
+            self._task.initialize(
+                get_examples=get_examples,
+                nlp=nlp,
+                **kwargs,
+            )
 
     def to_bytes(self, *, exclude: Tuple[str] = cast(Tuple[str], tuple())) -> bytes:
         """Serialize the LLMWrapper to a bytestring.
--- a/spacy_llm/tasks/ner.py
+++ b/spacy_llm/tasks/ner.py
@@ -18,14 +18,18 @@ _DEFAULT_NER_TEMPLATE_V2 = read_template("ner.v2")
 
 @registry.llm_tasks("spacy.NER.v1")
 def make_ner_task(
-    labels: str,
+    labels: str = "",
     examples: Optional[Callable[[], Iterable[Any]]] = None,
     normalizer: Optional[Callable[[str], str]] = None,
     alignment_mode: Literal["strict", "contract", "expand"] = "contract",  # noqa: F821
     case_sensitive_matching: bool = False,
     single_match: bool = False,
 ):
-    labels_list = split_labels(labels)
+    if labels:
+        labels_list = split_labels(labels)
+    else:
+        labels_list = []
+
     span_examples = (
         [SpanExample(**eg) for eg in examples()] if callable(examples) else examples
     )
@@ -42,7 +46,7 @@ def make_ner_task(
 
 @registry.llm_tasks("spacy.NER.v2")
 def make_ner_task_v2(
-    labels: Union[List[str], str],
+    labels: Union[List[str], str] = [],
     template: str = _DEFAULT_NER_TEMPLATE_V2,
     label_definitions: Optional[Dict[str, str]] = None,
     examples: ExamplesConfigType = None,
@@ -107,6 +111,36 @@ class NERTask(SpanTask):
             case_sensitive_matching=case_sensitive_matching,
             single_match=single_match,
         )
+
+    def initialize(
+        self,
+        get_examples: Callable[[], Iterable["Example"]],
+        nlp: Iterable["Example"],
+        labels: List[str] = None,
+        **kwargs: Any,
+    ):
+        """Initialize the NER task, by auto-discovering labels.
+
+        Labels can be set through, by order of precedence:
+
+        - the `[initialize]` section of the pipeline configuration
+        - the `labels` argument supplied to the task factory
+        - the labels found in the examples
+        """
+        examples = get_examples()
+
+        if not labels:
+            labels = list(self._label_dict.values())
+
+        if not labels:
+            label_set = set()
+
+            for eg in examples:
+                for ent in eg.reference.ents:
+                    label_set.add(ent.label_)
+            labels = list(label_set)
+
+        self._label_dict = {self._normalizer(label): label for label in labels}
 
     def assign_spans(
         self,
--- a/spacy_llm/tasks/spancat.py
+++ b/spacy_llm/tasks/spancat.py
@@ -41,7 +41,7 @@ def make_spancat_task(
 
 @registry.llm_tasks("spacy.SpanCat.v2")
 def make_spancat_task_v2(
-    labels: Union[List[str], str],
+    labels: Union[List[str], str] = [],
     template: str = _DEFAULT_SPANCAT_TEMPLATE_V2,
     label_definitions: Optional[Dict[str, str]] = None,
     examples: ExamplesConfigType = None,
@@ -68,7 +68,7 @@ def make_spancat_task_v2(
 class SpanCatTask(SpanTask):
     def __init__(
         self,
-        labels: List[str],
+        labels: List[str] = [],
         template: str = _DEFAULT_SPANCAT_TEMPLATE_V2,
         label_definitions: Optional[Dict[str, str]] = None,
         spans_key: str = "sc",
@@ -83,6 +83,7 @@ class SpanCatTask(SpanTask):
         """Default SpanCat task.
 
         labels (str): Comma-separated list of labels to pass to the template.
+            Leave empty to populate it at initialization time.
         spans_key (str): Key of the `Doc.spans` dict to save under.
         examples (Optional[Callable[[], Iterable[Any]]]): Optional callable that
             reads a file containing task examples for few-shot learning. If None is
@@ -122,3 +123,33 @@ class SpanCatTask(SpanTask):
             spans_key=self._spans_key,
             allow_overlap=True,
         )
+
+    def initialize(
+        self,
+        get_examples: Callable[[], Iterable["Example"]],
+        nlp: Iterable["Example"],
+        labels: List[str] = None,
+        **kwargs: Any,
+    ):
+        """Initialize the NER task, by auto-discovering labels.
+
+        Labels can be set through, by order of precedence:
+
+        - the `[initialize]` section of the pipeline configuration
+        - the `labels` argument supplied to the task factory
+        - the labels found in the examples
+        """
+        examples = get_examples()
+
+        if not labels:
+            labels = list(self._label_dict.values())
+
+        if not labels:
+            label_set = set()
+
+            for eg in examples:
+                for span in eg.reference.spans[self._spans_key]:
+                    label_set.add(span.label_)
+            labels = list(label_set)
+
+        self._label_dict = {self._normalizer(label): label for label in labels}


bdura:
@rmitsch I'd love your input at this point, before I move on to add label extraction for other tasks.

For instance, I've added support for supplying the labels in the `[initialize]` section; on the one hand it would align with what's done for other components, on the other hand it's a bit awkward since by design those would be supplied in the config directly... In general I tend to think that multiple ways of doing the same thing leads to more issues than advantages.

spacy_llm/pipeline/llm.py
```diff
@@ -1,11 +1,23 @@
...
+    Callable,
+    Dict,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Tuple,
+    Union,
+    cast,
+)
```
rmitsch, line 13: Minor: in line with spaCy we don't want imports to be broken up into multiple lines (seperate it into n lines if necessary).

spacy_llm/pipeline/llm.py
```diff
@@ -164,6 +176,19 @@ def _process_docs(self, docs: List[Doc]) -> List[Doc]:
 
         return final_docs
 
+    def initialize(
+        self,
+        get_examples: Callable[[], Iterable["Example"]],
+        nlp: Iterable["Example"],
+        **kwargs: Any,
+    ):
+        if isinstance(self._task, InitializableComponent):
```
rmitsch, line 185: `InitializableComponent` is a bit misleading to use for `Task`s, but it also doesn't make sense to redefine it. 
I'm confused by `nlp: Iterable["Example"]`, is this really intentational? (I know it's like that in `spacy.ty` too).

spacy_llm/tasks/ner.py
```diff
@@ -108,6 +112,36 @@ def __init__(
...
+        **kwargs: Any,
+    ):
+        """Initialize the NER task, by auto-discovering labels.
+
+        Labels can be set through, by order of precedence:
+
+        - the `[initialize]` section of the pipeline configuration
+        - the `labels` argument supplied to the task factory
+        - the labels found in the examples
+        """
```
rmitsch, line 129: Describe params.

spacy_llm/tasks/ner.py
```diff
@@ -108,6 +112,36 @@ def __init__(
             single_match=single_match,
         )
 
+    def initialize(
+        self,
+        get_examples: Callable[[], Iterable["Example"]],
+        nlp: Iterable["Example"],
+        labels: List[str] = [],
+        **kwargs: Any,
+    ):
```
rmitsch, line 121:
```suggestion
    ) -> None:
```

spacy_llm/pipeline/llm.py
```diff
@@ -164,6 +176,19 @@ def _process_docs(self, docs: List[Doc]) -> List[Doc]:
 
         return final_docs
 
+    def initialize(
+        self,
+        get_examples: Callable[[], Iterable["Example"]],
+        nlp: Iterable["Example"],
+        **kwargs: Any,
+    ):
```
rmitsch, line 184:
```suggestion
    ) -> None:
```

spacy_llm/tasks/spancat.py
```diff
@@ -122,3 +123,33 @@ def scorer(
...
             allow_overlap=True,
         )
+
+    def initialize(
+        self,
+        get_examples: Callable[[], Iterable["Example"]],
+        nlp: Iterable["Example"],
+        labels: List[str] = [],
+        **kwargs: Any,
+    ):
```
rmitsch, line 133:
```suggestion
    ) -> None:
```

spacy_llm/tasks/spancat.py
```diff
@@ -122,3 +123,33 @@ def scorer(
...
+        """Initialize the NER task, by auto-discovering labels.
+
+        Labels can be set through, by order of precedence:
+
+        - the `[initialize]` section of the pipeline configuration
+        - the `labels` argument supplied to the task factory
+        - the labels found in the examples
+        """
```
rmitsch, line 141: Describe params.

rmitsch (commented):
Looks reasonable to me. What do you think is awkward about this?

rmitsch:
Also: why not for textcat?

adrianeboyd:
If you add labels with `[initialize]` then this is only done once (say, reading from a file path that's only available initially) and then the component needs to save/load the labels with `to/from_bytes/disk`.

It is going to get confusing in terms of how to modify labels for an existing saved component. What would currently happen in practice, in terms of the order in which values are loaded/overwritten, would be component config -> initialize config (but only if run) -> deserialized config.

spacy_llm/pipeline/llm.py
bdura: True. I'll modify the pre-commit configuration in another PR, we've come up with a clean solution for that in Weasel.

spacy_llm/pipeline/llm.py
bdura: You're right, this is misleading. I can always artificially rename the `InitializableComponent`.

> I'm confused by nlp: Iterable["Example"], is this really intentational? (I know it's like that in spacy.ty too).

Me too, I meant to change it but got focused elsewhere. I'll also push a PR to spaCy.

bdura:
> say, reading from a file path that's only available initially

Aaah right, I forgot about that particular use case. Makes a lot of sense then, thanks

spacy_llm/pipeline/llm.py
bdura: [spaCy PR](https://github.com/explosion/spaCy/pull/12692).

bdura:
[spaCy's `validate_init_settings` function](https://github.com/explosion/spaCy/blob/master/spacy/schemas.py#L111) does not support `**kwargs: Any`. I changed the way the `initialize` method is exposed to curb that issue.

spacy_llm/tasks/spancat.py
```diff
@@ -83,6 +84,7 @@ def __init__(
         """Default SpanCat task.
 
         labels (str): Comma-separated list of labels to pass to the template.
+            Leave empty to populate it at initialization time.
```
rmitsch, line 87:
```suggestion
            Leave empty to populate it at initialization time (only if examples are provided).
```

spacy_llm/tasks/rel.py
```diff
@@ -58,7 +60,7 @@ def _preannotate(doc: Union[Doc, RELExample]) -> str:
 
 @registry.llm_tasks("spacy.REL.v1")
 def make_rel_task(
-    labels: Union[List[str], str],
+    labels: Union[List[str], str] = [],
```
rmitsch, line 63: Can you add a docstring for this as well?

spacy_llm/tasks/spancat.py
```diff
@@ -41,7 +42,7 @@ def make_spancat_task(
 
 @registry.llm_tasks("spacy.SpanCat.v2")
 def make_spancat_task_v2(
-    labels: Union[List[str], str],
+    labels: Union[List[str], str] = [],
```
rmitsch, line 45: Can you add a docstring for this as well?

spacy_llm/tasks/textcat.py
```diff
@@ -24,7 +25,7 @@ class TextCatExample(BaseModel):
 
 @registry.llm_tasks("spacy.TextCat.v1")
 def make_textcat_task(
-    labels: str,
+    labels: str = "",
```
rmitsch, line 28: Can you add a docstring for this as well?

spacy_llm/tasks/textcat.py
```diff
@@ -49,7 +50,7 @@ def make_textcat_task(
 
 @registry.llm_tasks("spacy.TextCat.v2")
 def make_textcat_task_v2(
-    labels: Union[List[str], str],
+    labels: Union[List[str], str] = [],
```
rmitsch, line 53: Can you add a docstring for this as well?

rmitsch (commented):
Great work, almost there. There's a typo in `def _check_rel_extention(cls)` - couldn't comment in the code (since when can you comment only on changed lines?).

rmitsch:
> If you add labels with [initialize] then this is only done once (say, reading from a file path that's only available initially) and then the component needs to save/load the labels with to/from_bytes/disk.

Ah, but we don't (de-)serialize the labels yet?

adrianeboyd:
Right, this PR would need to add that to `llm`.

spacy_llm/tasks/util/span.py
```diff
@@ -12,6 +14,15 @@
 class SpanTask:
     """Base class for Span-related tasks, eg NER and SpanCat."""
 
+    PLAIN_CONFIG_KEYS: List[str] = [
```
rmitsch, line 17: Assuming this shouldn't be accessed externally (also for other occurrences).
```suggestion
    _PLAIN_CONFIG_KEYS: List[str] = [
```

spacy_llm/tests/tasks/test_spancat.py
```diff
@@ -557,3 +557,29 @@ def test_spancat_init(noop_config):
...
+    task1._label_dict = labels
+
+    assert task1._label_dict == labels
+    assert task2._label_dict == dict()
+
+    b = nlp1.to_bytes()
+    nlp2.from_bytes(b)
+
+    assert task1._label_dict == labels
+    assert task2._label_dict == labels
```
rmitsch, line 585:
```suggestion
    assert task1._label_dict ==  assert task2._label_dict == labels
```

spacy_llm/tasks/util/span.py
bdura: 👍

spacy_llm/tasks/util/span.py
```diff
@@ -103,3 +114,62 @@ def parse_responses(
...
+            return srsly.json_dumps(plain)
+
+        def serialize_examples():
+            if self._examples is None:
+                return srsly.json_dumps(None)
+            examples = [eg.dict() for eg in self._examples]
+            return srsly.json_dumps(examples)
+
+        serialize = {
+            "plain": serialize_plain_config,
```
adrianeboyd, line 140: To make this more similar to existing methods, I'd suggest `cfg` instead of `plain`.

spacy_llm/pipeline/llm.py
```diff
@@ -238,4 +249,16 @@ def from_disk(
         exclude (Tuple): Names of properties to exclude from deserialization.
         RETURNS (LLMWrapper): Modified LLMWrapper instance.
         """
+
+        serialize = {}
+
+        task = self._task
+        backend = self._backend
```
rmitsch, line 256: Why introduce aliases here?

spacy_llm/tasks/util/span.py
```diff
@@ -11,10 +13,53 @@
 from .parsing import find_substrings
 
 
+def serialize_cfg(task: "SpanTask", cfg_keys: List[str]) -> str:
```
rmitsch, line 16: Why not as class method(s)?

rmitsch (commented):
Thanks! Almost there :pray:

spacy_llm/pipeline/llm.py
bdura: I was trying to fool the static type checker, but that was no remedy... I'll remove those, thanks.

spacy_llm/tasks/util/span.py
bdura: Good point.

spacy_llm/tasks/util/span.py
```diff
@@ -1,20 +1,19 @@
...
 from .examples import SpanExample
 from .parsing import find_substrings
+from .serialization import SerializableTask
 
 
-class SpanTask:
+class SpanTask(SerializableTask[SpanExample]):
     """Base class for Span-related tasks, eg NER and SpanCat."""
 
+    _Example = SpanExample
```
rmitsch, line 16: Why do you redefine `_Example` here? Should the injected `Generic` type not be sufficient for type hinting purposes?

spacy_llm/tasks/util/serialization.py
```diff
@@ -0,0 +1,125 @@
...
+from pydantic import BaseModel
+from spacy import util
+
+ExampleType = TypeVar("ExampleType", bound=BaseModel)
+
+
+class SerializableTask(Generic[ExampleType]):
+    """A task that can be serialized and deserialized."""
+
+    _CFG_KEYS: List[str]
```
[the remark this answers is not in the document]
bdura: Good point, thanks.

spacy_llm/tasks/util/span.py
bdura: (and `_Example.parse_obj`, which is the actual constraint)

spacy_llm/tasks/util/serialization.py
```diff
@@ -0,0 +1,135 @@
...
+        exclude: Tuple[str] = cast(Tuple[str], tuple()),
+    ) -> None:
+        """Serialize the task to disk.
+
+        path (Path): A path (currently unused).
+        exclude (Tuple): Names of properties to exclude from serialization.
+        """
+
+        serialize = {
+            "cfg.json": lambda p: p.write_text(self.serialize_cfg()),
```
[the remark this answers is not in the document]
bdura: Thanks @adrianeboyd, that makes a lot of sense. I don't think the data will get too complicated, although examples might get bigger.

@rmitsch, any thought about `msgpack`?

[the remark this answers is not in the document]
rmitsch: I see you already switched to `msgpack` meanwhile - no objections.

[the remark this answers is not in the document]
rmitsch: I don't expect that there will be a ton of examples, as it'll be mostly few-shot use cases, I assume.

rmitsch (approved):
Great work! :tada: I think this is ready to go (if @adrianeboyd doesn't have further comments).

PR #153, opened 2023-05-31
specify config_path as Union[str, Path]

svlandeg:
## Description
These scripts are in the example usages dir, so it'd be good if they reflect the actual correct type from the `assemble` functionality in case users are copy/pasting etc.

### Types of change
fix

## Checklist
  - all tests requiring a GPU


adrianeboyd:
I don't think `typer` supports `Union` types and `Path` works out fine in practice for this kind of script? spacy also uses `Path` for these kinds of args.

rmitsch:
> I don't think typer supports `Union` types...

It doesn't, unfortunately.

svlandeg:
Ok, fair enough. Closing :-)

PR #157, opened 2023-06-01
Add Anthropic REST backend

ljvmiranda921:
Adds a REST backend for Anthropic's Claude API. [GH Action run should be passing](https://github.com/explosion/spacy-llm/actions/runs/5250545233).

## Description

Can't test these yet because I don't have API access yet. But I have two design related questions:

1. Anthropic requires a [different formatting for their prompts](https://console.anthropic.com/docs/api#prompt-formatting), I'm doing this formatting internally rather than in a jinja template somewhere. I'm not sure how this will affect the accuracy though.
2. Unlike OpenAI's API, Anthropic doesn't take in a batch set of prompts. So I'm doing something like:

```python
for prompt in prompts:
    _request({"prompt": prompt})
```

Which I don't think is ideal. What's the best practice here?

### Types of change

- New REST backend

## Checklist

  - all tests requiring a GPU


rmitsch:
> Which I don't think is idea. What's the best practice here?

If they don't support batching, it's either what you're doing or writing a batch prompt, which is a pain (and brittle). So let's stick with individual prompt execution for now, if there's no batching.

spacy_llm/tests/backends/test_anthropic.py
```diff
@@ -0,0 +1,23 @@
+# mypy: ignore-errors
+import pytest
+from typing import Iterable
+
+from spacy.tokens import Doc
+
+from ...registry import registry
+from ..compat import has_anthropic_key
```
svlandeg, line 8: Only use the relative imports in the test suite if it's other modules from the test suite, which is the case for `compat` here. For `registry`, we should import explicitely from `spacy_llm.registry`.

spacy_llm/backends/rest/anthropic.py
```diff
@@ -0,0 +1,121 @@
...
+                    raise ValueError(f"API call failed: {response}.")
+                else:
+                    assert isinstance(prompts, Sized)
+                    return {"error": [srsly.json_dumps(response)] * len(prompts)}
+            return response
+
+        responses = [
+            _request({"prompt": f"{Speaker.HUMAN} {prompt}{Speaker.ASST}"})
+            for prompt in prompts
+        ]
```
svlandeg, line 112: Agreed with Raphael that simulating the batching like you're doing, makes most sense. But perhaps you can leave a comment in the code here explaining that batching is not currently available for this endpoint, hence this solution.

svlandeg (commented):
This will be good to have!

ljvmiranda921:
Semi-ready for review but would probably wait for the API key before merging

ljvmiranda921:
Moving this to unassigned for now. I'll get back to this once we get access to API keys so I can extensively test the feature. :)

rmitsch:
@ljvmiranda921 What's the status on this? Looks review-ready for me :slightly_smiling_face:

ljvmiranda921:
This is also review ready!

rmitsch:
@ljvmiranda921 Will merge after you resolved the conflicts :slightly_smiling_face:

PR #158, opened 2023-06-01
Use equality operator in tests

vin-ivar:
## Description
Most of the test suite fails on some machines when checking types - a comparison with `is` consistently returns `False` even if the types are the same. This PR fixes that. 

### Types of change
Bug fix.

## Checklist

  - all tests requiring a GPU


rmitsch (approved):
Thanks for catching and fixing this!

This works for me locally to, and the tests pass as well. I guess there's no harm in merging this. We can still evaluate the type checking more generally at a later point.

svlandeg (approved):
Thanks!

PR #163, opened 2023-06-07
feat: create examples on init

bdura:
## Description

This PR adds a mechanism to gather examples at init time.

### Types of change

This is a feature.

## Checklist

  - all tests requiring a GPU


bdura:
It's missing example extraction for other tasks, but I wanted to let you have a look before I took a break.

spacy_llm/tasks/ner.py
```diff
@@ -198,3 +204,21 @@ def scorer(
         examples: Iterable[Example],
     ) -> Dict[str, Any]:
         return get_ner_prf(examples)
+
+    def create_examples(
+        self,
+        examples: List[Example],
+    ) -> List[SpanExample]:
+        """Create NER examples from spaCy examples."""
```
rmitsch, line 212: - Incomplete docstring.
- `create_examples(List[Example]) -> List[SpanExample]` reads a bit confusing to me due to the naming conflict between spaCy's `Example` and `spacy-llm`'s `Example` type(s). Perhaps we either rename this function or the `Example` types in `spacy-llm`?

rmitsch: I'm in favor of the latter. `Demonstration` maybe? CC @svlandeg

spacy_llm/tests/tasks/test_textcat.py
```diff
@@ -759,6 +763,10 @@ def test_textcat_init(noop_config, init_from_config: bool):
     else:
         target = set()
     assert set(task._label_dict.values()) == target
+    assert not task._examples
+
+    # This is super hacky... but it works for now.
+    nlp.config["initialize"]["components"]["llm"] = {"gather_examples": gather_examples}
```
rmitsch, line 769: I feel this is ok for the tests.

rmitsch (commented):
Looks good overall. My main issue with this is the overloaded use of "Example" - we had that before, but this PR makes it clearer how confusing it can be to talk about spaCy and `spacy-llm` examples in the same context. As mentioned in a comment I advocate renaming the `spacy-llm` "Example" to "Demonstration" or something similar. Would like to hear about this from others.

spacy_llm/tasks/ner.py
svlandeg: We don't have an actual `Example` type in `spacy-llm`, right? Only things like `RELExample` and `SpanExample`?

I do agree that `create_examples` is confusing here. Perhaps `create_prompt_examples`?

spacy_llm/tasks/ner.py
```diff
@@ -183,6 +186,9 @@ def initialize(
                     label_set.add(ent.label_)
             labels = list(label_set)
 
+        if gather_examples:
+            self._examples = self.create_examples(examples)
```
svlandeg, line 190: Maybe also here use a more explicit name, e.g. `self._prompt_examples`?

spacy_llm/tasks/ner.py
```diff
@@ -168,9 +170,10 @@ def initialize(
             for initialization.
         nlp (Language): Language instance.
         labels (List[str]): Optional list of labels.
+        gather_examples (bool): Whether to gather examples for the task. False by default.
         """
 
-        examples = get_examples()
+        examples = list(get_examples())
```
svlandeg, line 176: Instead of putting all the `examples` in a list, we can rewrite the following lines to only loop through all `examples` once, and for each one add to the `label_set` if we need to deduce labels, and create the prompt_example and add it to the internal list if `gather_examples` (`gather_prompt_examples`?) is `True`.

spacy_llm/tasks/ner.py
rmitsch: Ah, that's right. `create_prompt_examples` is good with me.

spacy_llm/tasks/ner.py
rmitsch: Sounds good.

spacy_llm/tasks/ner.py
rmitsch: Agreed.

spacy_llm/tasks/util/serialization.py
```diff
@@ -39,20 +39,20 @@ def set_cfg(self, cfg: Dict[str, Any]) -> None:
         for key, value in cfg.items():
             setattr(self, key, value)
 
-    def get_examples(self) -> Optional[List[Dict[str, Any]]]:
+    def get_prompt_examples(self) -> Optional[List[Dict[str, Any]]]:
```
svlandeg, line 42: We shouldn't have called this `get_examples` in https://github.com/explosion/spacy-llm/pull/152 because typically we have a `get_examples: Callable[[], Iterable["Example"]]` in `initialize()` which works on spaCy's `Example` objects. So I suggest to be explicit everytime we mean "prompt" example, to avoid confusion.

spacy_llm/tasks/ner.py
```diff
@@ -166,21 +168,26 @@ def initialize(
             for initialization.
         nlp (Language): Language instance.
         labels (List[str]): Optional list of labels.
+        infer_prompt_examples (int): How many prompt examples to infer from the Example objects. 0 by default.
```
svlandeg, line 171: I suggest to make this an `int` instead of the `bool` we had before, so users can give a larger file with examples (to infer labels from) but only have like 2 in the prompt.

svlandeg: And to have any negative number (e.g. `-1` signals "all examples")

svlandeg:
Note: still missing docs (let's go through review first)

spacy_llm/tasks/ner.py
```diff
@@ -166,21 +168,30 @@ def initialize(
...
-            labels = list(label_set)
-
+                    labels.append(ent.label_)
+            if (
+                infer_prompt_examples < 0
+                or len(self._prompt_examples) < infer_prompt_examples
+            ):
+                self._prompt_examples.append(self._create_prompt_example(eg))
+
+        labels = list(set(labels))
```
rmitsch, line 194: Why the list conversion?

rmitsch (approved):
LGTM!

spacy_llm/tasks/ner.py
svlandeg: So in the original version of this code we had a `label_set` variable that was used during label inference, to avoid duplicates, and this was then transformed to a list and put into `labels`. I figured that whichever way the labels come in, we want to avoid having duplicates, and I'm assuming that the list of labels won't be ridiculously long for this operation to become cumbersome.

rmitsch: Right, I mean we don't need this to be a `list`? Like
```python
self._label_dict = {self._normalizer(label): label for label in set(labels)}
```
would suffice.

rmitsch:
Retracted the approval for clarity because docs are still missing. Code is good for me.

spacy_llm/tasks/ner.py
svlandeg: Right! I somehow thought that we were storing this list internally. I'll make the change (and add the docs as well).

svlandeg (commented):
Hm, we have the bump all the versions of the tasks because we're adding functionality that wasn't previously available, so otherwise it's not clear from the docs when users can actually do this.

spacy_llm/tasks/lemma.py
```diff
@@ -114,5 +126,10 @@ def _cfg_keys(self) -> List[str]:
         return ["_template"]
 
     @property
-    def _Example(self) -> Type[ExampleType]:
+    def _Example(self) -> Type[LemmaExample]:
         return LemmaExample
+
+    def _create_prompt_example(self, example: Example) -> LemmaExample:
+        """Create a lemma prompt example from a spaCy example."""
+        lemma_dict = [{t.text: t.lemma_ for t in example.reference}]
```
svlandeg, line 134:
```suggestion
        lemma_dict = [{t.text: t.lemma_} for t in example.reference]
```

rmitsch:
I'm not super happy about the name `infer_prompt_examples` actually, sounds like a flag. `n_prompt_examples_to_infer` maybe? It's verbose, I know, but more accurate.

spacy_llm/tasks/ner.py
```diff
@@ -166,22 +168,29 @@ def initialize(
             for initialization.
         nlp (Language): Language instance.
         labels (List[str]): Optional list of labels.
+        infer_prompt_examples (int): How many prompt examples to infer from the Example objects.
```
rmitsch, line 171: The description doesn't feel clear enough to me. Maybe 
> The maximal number of prompt examples to inject into prompts.
?

rmitsch:
More generally I'm a bit unsure about this because we offer two way to do one thing (inject examples into the prompt):
1. Create an examples file and inject it into a prompt using the `spacy.FileReader.v1`.
2. Initialize the component with `Example` instances, set `infer_prompt_examples != 0`.

I feel like this may be confusing. I do see the value though.

spacy_llm/tasks/ner.py
rmitsch: It's ok, actually. The alternative description doesn't add much.

rmitsch:
> This will be a hassle because we'll have the legacy versions, or we might just drop support for the older versions as we go to the next minor version...

I'm fine with just dropping all legacy versions and always using `.v1` as long as we are <`1.0.0` (as long as we version afterwards).

rmitsch:
@svlandeg I think this is good to go except for the readme updates. I can wrap this up tomorrow, but feel free to do so today.

rmitsch:
Updated the readme. Ready for review.

svlandeg (approved):
Thanks for finishing this up Raphael! This PR's been a true team effort :-)

PR #166, opened 2023-06-08
Fix issue with blank prompts

vin-ivar:
## Description
I'm not sure if this is an issue for every kind of model, but at least with `text-davinci-003`, loading from a cache fails because you're still attempting to parse the non-cached examples in a batch - if this list is empty, the model receives a blank prompt and exits, this should fix that by explicitly checking.

### Types of change
Bug fix.

## Checklist

  - all tests requiring a GPU


rmitsch (commented):
Thanks for the fix! Can you clean up the pre-commit errors?

rmitsch (approved):

PR #176, opened 2023-06-19
Refactor backend -> model

rmitsch:
## Description

Refactor to switch from a _backend_ to a _model_ paradigm.
Todos:
- Readme updates
- Update Streamlit demo

## Checklist

  - all tests requiring a GPU


README.md
```diff
@@ -68,7 +68,7 @@ factory = "llm"
 labels = ["COMPLIMENT", "INSULT"]
 
 [components.llm.backend]
```
svlandeg, line 70: We need to make sure the readme examples are updated correctly, many of them still contain references to `[components.llm.backend]`

spacy_llm/models/hf/dolly.py
```diff
@@ -34,9 +24,17 @@ def __call__(self, prompts: Iterable[str]) -> Iterable[str]:  # type: ignore[ove
...
         ]
 
+    @property
+    def model_name(self) -> str:
+        return f"databricks/dolly-{self._variant}"
+
+    @staticmethod
+    def get_model_variants() -> Iterable[str]:
+        return ["v2-3b", "v2-7b", "v2-12b"]
+
```
svlandeg, line 34: I like the idea here and the fact that we'd type this as a `Literal` in the `dolly_hf` function, but it feels to me that `model` or `model_name`, including the "dolly-" part, is much more intuitive to users, and readable in the config, than only specifying the variant suffix.

spacy_llm/models/rest/base.py
```diff
@@ -17,18 +16,27 @@ def has(cls, item: int):
...
 
 
-class Backend(abc.ABC):
+class Model(abc.ABC):
     """Queries LLMs via their REST APIs."""
 
+    DEFAULT_STRICT = True
+    DEFAULT_MAX_TRIES = 5
+    DEFAULT_INTERVAL = 1.0
+    DEFAULT_MAX_REQUEST_TIME = 30
```
svlandeg, line 25: Having this in ALL_CAPS would signal to me that they are invariant, but they are actually variables on the class that can be overwritten by subclasses, right? So basically they are invariant per subclass, but can differ between classes?

README.md
rmitsch: Yes, I deliberately haven't updated the readme yet as I wasn't sure if we stick with the proposed design.

spacy_llm/models/hf/dolly.py
rmitsch: So you'd keep the `variant` concept, but use the fully qualified name like `databricks/dolly-v2-3b`?
I'd like not to name this `model` because I expect there to be some confusion in instantiating a `Model` class with a `model` argument. Maybe `name`?

spacy_llm/models/rest/base.py
rmitsch: It's not something I'd expect to happen, but theoretically yes (still disappointed there is no `const` in Python). Do you prefer them to be lowercase?

spacy_llm/models/hf/dolly.py
svlandeg: I personally think this would be nicest:
```
name = "dolly-v2-3b"
```
It's sort of a consensus, dropping the "databricks" prefix but not the "dolly" part so it's still readable by itself. And then I'd still have the `Literal` which is nice.

svlandeg: Maybe internally it can be `model_name`?

spacy_llm/models/rest/base.py
svlandeg: Hm meh, I don't actually know, it's this in-between case. Maybe we just leave as-is.

spacy_llm/models/hf/dolly.py
rmitsch: Updated in https://github.com/explosion/spacy-llm/pull/176/commits/f78adfcf8f1cee6694253832b2f4892ce2cc0203.

README.md
```diff
@@ -11,7 +11,7 @@ This package integrates Large Language Models (LLMs) into [spaCy](https://spacy.
 - Serializable `llm` **component** to integrate prompts into your pipeline
 - **Modular functions** to define the [**task**](#tasks) (prompting and parsing) and [**backend**](#backends) (model to use)
 - Support for **hosted APIs** and self-hosted **open-source models**
-- Integration with [`MiniChain`](https://github.com/srush/MiniChain) and [`LangChain`](https://github.com/hwchase17/langchain)
```
svlandeg, line 14: Do I understand it correctly that MiniChain just isn't used much?

spacy_llm/models/hf/base.py
```diff
@@ -4,41 +4,40 @@
...
+        self._hf_model_id = f"{self.get_hf_account()}/{name}"
         self._config_init, self._config_run = self.compile_default_configs()
         if config_init:
             self._config_init = {**self._config_init, **config_init}
         if config_run:
             self._config_run = {**self._config_run, **config_run}
 
         # Init HF model.
-        HuggingFaceBackend.check_installation()
-        self.check_model()
```
svlandeg, line 41: Why remove this `check_model` call & functionality?

spacy_llm/models/hf/openllama.py
```diff
@@ -75,20 +77,22 @@ def compile_default_configs() -> Tuple[Dict[str, Any], Dict[str, Any]]:
...
-def backend_openllama_hf(
-    model: str,
+@registry.llm_models("spacy.OpenLLaMA.v1")
+def openllama_hf(
+    name: Literal[
+        "open_llama_3b_350bt_preview",
+        "open_llama_3b_600bt_preview",
+        "open_llama_7b_400bt_preview",
+        "open_llama_7b_600bt_preview",
+    ],  # noqa: F722
```
svlandeg, line 87: Is there a way to avoid hard-coding these twice, both here and in `get_model_names`?

spacy_llm/models/integration/langchain.py
```diff
@@ -0,0 +1,73 @@
...
+                return Remote(
+                    integration=type_to_cls_dict[langchain_class_id](
+                        model_name=model, **config
+                    ),
+                    query=query_langchain() if query is None else query,
+                )
+            except ImportError as ex:
+                raise ValueError(
+                    "Failed to instantiate LangChain class. Ensure all necessary dependencies are installed."
+                ) from ex
```
svlandeg, line 62: This probably should error/exit, but warn instead? Because you might just want to use `spacy-llm` without LangChain but then if this breaks, you're blocked.

spacy_llm/models/hf/base.py
rmitsch: I considered it redundant as we only offer per-model instantiation now. On the other hand, if users instantiate a HF model in the code, it might still be helpful (because then the model name isn't hardcoded)...what do you think?

README.md
rmitsch: Yes. We initially used MiniChain as our default backend due to its simplicity. Now that we have our REST-based backend, I don't see an advantage to keeping MiniChain.

spacy_llm/models/integration/langchain.py
rmitsch: `register_langchain_models()` is only called [if `langchain` is available](https://github.com/explosion/spacy-llm/pull/176/files#diff-b461f0886b68b6bbf5bda73e7feebf4f6667fc7daa58a14cb4b9a62cb9b4196cR70).

spacy_llm/models/hf/base.py
rmitsch: I'm leaning towards pushing users to use the registered functions (they can still call e. g. `openai_gpt_4()` if they don't want to use the config system) and not offering support for instantiating `HuggingFaceModel` instances directly.

spacy_llm/models/hf/openllama.py
rmitsch: A suggestion for that: https://github.com/explosion/spacy-llm/pull/176/commits/b068e1d1542e5a0c8f141469d1bb9a284ba144ce.

spacy_llm/models/hf/base.py
```diff
@@ -4,41 +4,40 @@
...
         query (Callable[[Any, Iterable[_PromptType]], Iterable[_ResponseType]]): Callable executing LLM prompts when
             supplied with the `integration` object.
-        model (str): Name of HF model to load.
+        name (str): Name of HF model to load (without account name).
         config_init (Optional[Dict[str, Any]]): HF config for initializing the model.
         config_run (Optional[Dict[str, Any]]): HF config for running the model.
         inference_config (Dict[Any, Any]): HF config for model run.
         """
-        self._model_name = model
+        self._hf_model_id = f"{self.get_hf_account()}/{name}"
```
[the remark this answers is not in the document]
rmitsch: Done in https://github.com/explosion/spacy-llm/pull/176/commits/b068e1d1542e5a0c8f141469d1bb9a284ba144ce.

spacy_llm/models/hf/base.py
svlandeg: There's still a list of names that are supported though, right? I'm thinking of use-cases where someone writes in the config
```
[components.llm.model]
@llm_models = "spacy.gpt-3-5.v1"
gpt-3.5-turbot
```

spacy_llm/models/integration/langchain.py
svlandeg: I realise that, but even then. Imagine having LangChain in your environment, but you're not actually using it, and these register commands break. Then you can't continue with any functionality in `spacy-llm`. In general, we should really only error if we're 100% sure that there's no further usage possible.

rmitsch: Fair point. Warning instead of raising in https://github.com/explosion/spacy-llm/pull/176/commits/05c4ee2e4ca4efdc9f2a62f5f70038dbdc4c4d3a.

spacy_llm/models/hf/base.py
svlandeg: I thought it was fine to have this as a Python utility, that directly queries the internals instead of having to manually look up in the docs? I can imagine the Python function to be useful to developers.

rmitsch: Reintroduced in https://github.com/explosion/spacy-llm/pull/176/commits/70fa93286f97ad351371e9a986a209248664f331.

rmitsch:
Apologies for the 11th-hour changes, but I took the opportunity to standardize the model classes a bit more in https://github.com/explosion/spacy-llm/pull/176/commits/70fa93286f97ad351371e9a986a209248664f331. I also considered adding another base model class, but that would have had rather little in advantages like code reuse vs. dealing with another level of abstraction.

This commit doesn't change a lot in terms of the API, it just does away with some unnecessary abstractions and standardizes the HF/LangChain/REST model API to some degree.

spacy_llm/models/hf/base.py
```diff
@@ -48,34 +52,45 @@ def __call__(self, prompts: Iterable[_PromptType]) -> Iterable[_ResponseType]:
         RETURNS (Iterable[_ResponseType]): API responses.
         """
 
-    @property
+    def _check_model(self) -> None:
+        """Checks whether model is supported. Raises if it isn't."""
+        if (
+            self._name not in self.get_model_names()
+            and self._name.replace(f"{self.get_hf_account()}/", "")
+            not in self.get_model_names()
```
svlandeg, line 60: At this point, shouldn't the name always include the prefix? So why the two checks?
If we feel this gets too confusing, we can also consider storing two variables internally, the `hf_account` and the `model_name` and never merge them up until we actually use the request to the HF API.

spacy_llm/models/hf/base.py
```diff
@@ -48,34 +52,45 @@ def __call__(self, prompts: Iterable[_PromptType]) -> Iterable[_ResponseType]:
...
+        RETURNS (Tuple[str]): Names of supported models.
+        """
+        return tuple(str(arg) for arg in cls.MODEL_NAMES.__args__)  # type: ignore[attr-defined]
+
+    @staticmethod
     @abc.abstractmethod
-    def supported_models(self) -> Iterable[str]:
-        """Get names of supported models.
-        RETURNS (Iterable[str]): Names of supported models.
+    def get_hf_account() -> str:
```
svlandeg, line 75: We should probably make this a property?
Something like 
```
@property
@abstractmethod
def hf_account(self):
```

spacy_llm/models/hf/base.py
```diff
@@ -48,34 +52,45 @@ def __call__(self, prompts: Iterable[_PromptType]) -> Iterable[_ResponseType]:
...
+        """Names of HF account for this model.
```
svlandeg, line 76:
```suggestion
        """Name of HF account for this model.
```

svlandeg:
Is there a way to catch `<=0.3.0` configs that refer to the old "backend" and omit some kind of warning about incompatibility, and refer to some sort of "migration" documentation from 0.3.x to 0.4.x?

usage_examples/README.md
```diff
@@ -97,27 +97,18 @@ You can check sample tasks for Named Entity Recognition and text categorization
...
 
-## Using LangChain, MiniChain and other integrated third-party prompting libraries
+## Using LangChain and other integrated third-party prompting libraries
 
 `spacy-llm` integrates bindings to a number of libraries centered on prompt management and LLM usage to allow users
-to leverage their functionality in their spaCy workflows. This currently includes
-
-- [LangChain](https://github.com/hwchase17/langchain)
-- [MiniChain](https://github.com/srush/MiniChain)
+to leverage their functionality in their spaCy workflows. This currently includes [LangChain](https://github.com/hwchase17/langchain)
```
svlandeg, line 104: I would rephrase and say that `spacy-llm` allows integrations with external libraries, and that LangChain is one built-in example.

svlandeg (commented):
Mostly looks good to me! Let's start wrapping this up? Raphael can you update the readme & double check the usage examples, and maybe write a short "migration" guide we can link to?

spacy_llm/models/hf/base.py
rmitsch: Done in https://github.com/explosion/spacy-llm/pull/176/commits/8d6ae3fa2c6246ad18554862fe586a6a3c1c2b28.

spacy_llm/models/hf/base.py
rmitsch: Fixed in https://github.com/explosion/spacy-llm/pull/176/commits/8d6ae3fa2c6246ad18554862fe586a6a3c1c2b28.

rmitsch:
> Is there a way to catch <=0.3.0 configs that refer to the old "backend" and omit some kind of warning about incompatibility, and refer to some sort of "migration" documentation from 0.3.x to 0.4.x?

Hm, only upstream, I think. As soon as the config is dropped into the pipeline, `confection` parses it and cries foul because `backend` is an illegal argument (and `model` is missing). I don't think we can do anything in `spacy-llm` about this.

svlandeg:
> > Is there a way to catch <=0.3.0 configs that refer to the old "backend" and omit some kind of warning about incompatibility, and refer to some sort of "migration" documentation from 0.3.x to 0.4.x?
> 
> Hm, only upstream, I think. As soon as the config is dropped into the pipeline, `confection` parses it and cries foul because `backend` is an illegal argument (and `model` is missing). I don't think we can do anything in `spacy-llm` about this.

We can't try to catch those specific errors?

rmitsch:
> We can't try to catch those specific errors from confection, to provide a more specific/helpful warning in spacy-llm for those specific cases? Like if there's an llm.backend config section - couldn't we even check that up-front in some cases?

I wouldn't know where to. Maybe I misunderstand something, but IMO the first thing that happens is that `@Language.factory` is run, including the config validation. I don't see how we can preempt or even catch that?

rmitsch:
> Mostly looks good to me! Let's start wrapping this up? Raphael can you update the readme & double check the usage examples, and maybe write a short "migration" guide we can link to?

Added migration guide and one last adjustment (`name` has to be specified explicitly for LangChain models now outside of the config dict to be more consistent with other models) in https://github.com/explosion/spacy-llm/pull/176/commits/502697fa1aa2e720ab6bea43abbbcf467527712d.

Running one last round of tests currently.

svlandeg:
> > We can't try to catch those specific errors from confection, to provide a more specific/helpful warning in spacy-llm for those specific cases? Like if there's an llm.backend config section - couldn't we even check that up-front in some cases?
> 
> I wouldn't know where to. Maybe I misunderstand something, but IMO the first thing that happens is that `@Language.factory` is run, including the config validation. I don't see how we can precede that or even catch that the resulting error?
> 
> [Edit] Well, I guess we can build our own `@Language.factory` that essentially calls spaCy's `@Language.factory`, but does those checks before. Not sure we really want that.

Yea, I guess you're right. We could make it a "FAQ" in the discussions forum so at least people will (hopefully) easily find the solution when they get that specific error?

usage_examples/README.md
```diff
@@ -100,7 +100,7 @@ in the `spacy_llm/tasks/` directory. We also recommend checking out the
 ## Using LangChain and other integrated third-party prompting libraries
 
 `spacy-llm` integrates bindings to a number of libraries centered on prompt management and LLM usage to allow users
```
svlandeg, line 102: I would still update this bit "integrates a number of libraries" as we really only have one, now.

README.md
```diff
@@ -761,22 +756,22 @@ export OPENAI_API_ORG="org-..."
...
-| `strict`    | `bool`           | `True`  | If `True`, raises an error if the LLM API returns a malformed response. Otherwise, return the error responses as is. |
-| `max_tries` | `int`            | `3`     | Max. number of tries for API request.                                                                                |
-| `timeout`   | `int`            | `30`    | Timeout for API request in seconds.                                                                                  |
+| Argument    | Type             | Default | Description                                                                                                                                               |
+| ----------- | ---------------- | ------- | --------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `name`      | `str`            |         | Model name. This might be any of the supported variants for this particular model (e. g. "gpt-3-5-turbo" or "gpt-3.5-turbo-16k" for ""spacy.gpt-3-5.v1"). |
```
svlandeg, line 770:
```suggestion
| `name`      | `str`            |         | Model name. This might be any of the supported variants for this particular model (e. g. "gpt-3-5-turbo" or "gpt-3.5-turbo-16k" for "spacy.gpt-3-5.v1"). |
```

README.md
```diff
@@ -968,18 +932,18 @@ python -m pip install "accelerate>=0.16.0,<1.0"
 Example config block:
 
 ```ini
-[components.llm.backend]
-@llm_backends = "spacy.OpenLLaMaHF.v1"
-model = "openlm-research/open_llama_3b_350bt_preview"
+[components.llm.model]
+@llm_models = "spacy.OpenLLaMAHF.v1"
```
svlandeg, line 936:
```suggestion
@llm_models = "spacy.OpenLLaMA.v1"
```
(right?)

migration_guide.md
```diff
@@ -0,0 +1,85 @@
...
+  <summary>0.3.x to 0.4.x</summary>
+
+## `0.3.x` to `0.4.x`
+
+`0.4.x` significantly refactors the code to make it more robust and the config more intuitive. 0.4.0 changes the config 
+paradigm from backend- to model-centric. This is reflected in the external API in a different config structure.
+
+Remember that there are three different types of models: the first uses the native REST implementation to communicate
+with hosted LLMs, the second builds on HuggingFace's `transformers` model to run models locally and the third leverages
+`langchain` to operate on hosted or local models. While they config for all three is rather similar (especially in
```
svlandeg, line 13:
```suggestion
`langchain` to operate on hosted or local models. While the config for all three is rather similar (especially in 
```

migration_guide.md
```diff
@@ -0,0 +1,85 @@
...
+`0.4.x` significantly refactors the code to make it more robust and the config more intuitive. 0.4.0 changes the config 
+paradigm from backend- to model-centric. This is reflected in the external API in a different config structure.
+
+Remember that there are three different types of models: the first uses the native REST implementation to communicate
+with hosted LLMs, the second builds on HuggingFace's `transformers` model to run models locally and the third leverages
+`langchain` to operate on hosted or local models. While they config for all three is rather similar (especially in 
+0.4.x), there are differences in how these models have to be configured. We show how to migrate your config from 0.3.x
+to 0.4.x for each of these model types.
+
+For all model types:
```
svlandeg, line 17:
```suggestion
### All model types 
```
for more focus on this part?

migration_guide.md
```diff
@@ -0,0 +1,85 @@
...
+paradigm from backend- to model-centric. This is reflected in the external API in a different config structure.
+
+Remember that there are three different types of models: the first uses the native REST implementation to communicate
+with hosted LLMs, the second builds on HuggingFace's `transformers` model to run models locally and the third leverages
+`langchain` to operate on hosted or local models. While they config for all three is rather similar (especially in 
+0.4.x), there are differences in how these models have to be configured. We show how to migrate your config from 0.3.x
+to 0.4.x for each of these model types.
+
+For all model types: 
+- The registry name has changed - instead of `llm_backends`, use `llm_models`.
```
svlandeg, line 18:
```suggestion
- The registry name has changed - instead of `@llm_backends`, use `@llm_models`.
```
I personally find that more clear, but I don't care strongly either way.

migration_guide.md
```diff
@@ -0,0 +1,85 @@
...
+@llm_models = "spacy.gpt-3-5.v1"
+name = "gpt-3-5-turbo"
+config = {"temperature": 0.3}
+```
+Note that the factory function (marked with `@`) refers to the name of the model. Variants of the same model can be 
+specified with the `name` attribute - for `gpt-3.5` this could be `"gpt-3-5-turbo"` or `"gpt-3-5-turbo-16k"`.
+
+### Models using HuggingFace
+
+On top of the changes described in the section above, HF models like `spacy.Dolly.v1` now accept `config_init` and
```
svlandeg, line 45: "above" is confusing here, because the section above is about the REST API. I would just remove this subsection, and start with "HF models like ..."

migration_guide.md
```diff
@@ -0,0 +1,85 @@
...
+@llm_backends = "spacy.Dolly_HF.v1"
+model = "databricks/dolly-v2-3b"
+config = {}
+```
+In `0.4.x`:
+```ini
+[components.llm.model]
+@llm_models = "spacy.Dolly.v1"
+name = "dolly-v2-3b"  # or databricks/dolly-v2-3b - the prefix is optional
+config_init = {}  # Arguments passed to HF model at initalization time
```
svlandeg, line 60:
```suggestion
config_init = {}  # Arguments passed to HF model at initialization time
```

spacy_llm/models/rest/cohere/registry.py
```diff
@@ -1,31 +1,26 @@
...
-    "Anthropic": anthropic.AnthropicBackend,
-    "NoOp": noop.NoOpBackend,
-}
 
-
-@registry.llm_backends("spacy.REST.v1")
-def backend_rest(
-    api: str,
+@registry.llm_models("spacy.command.v1")
+def cohere_command(
```
svlandeg, line 11: This is a case that I hadn't considered, that "command" is kind of a pretty generic name, though I guess everyone knows this is from Cohere? 😕

spacy_llm/models/rest/openai/registry.py
```diff
@@ -0,0 +1,374 @@
...
+        strict=strict,
+        max_tries=max_tries,
+        interval=interval,
+        max_request_time=max_request_time,
+    )
+
+
+@registry.llm_models("spacy.code-davinci.v1")
+def openai_code_davinci(
+    config: Dict[Any, Any] = SimpleFrozenDict(),
```
svlandeg, line 123: Wouldn't it be good to have the model name always as a required parameter, even if there's only a single model available? Just for consistency? Or do you think it'd be more of a burden than a help?

spacy_llm/pipeline/llm.py
```diff
@@ -34,10 +34,8 @@ class CacheConfigType(TypedDict):
     assigns=[],
     default_config={
         "task": None,
-        "backend": {
-            "@llm_backends": "spacy.REST.v1",
-            "api": "OpenAI",
-            "config": {"model": "gpt-3.5-turbo"},
+        "model": {
+            "@llm_models": "spacy.gpt-3-5.v1",
             "strict": True,
```
svlandeg, line 39: The model `name` isn't here? I would specify the default here for clarity.

svlandeg: This is making me reconsider whether the individual models should have a default value for their `name: Literal` setting 🤔 

(also fine by me to think about this in a follow up PR)

svlandeg (approved):
This is looking really good, thanks for pushing this one true Raphael! 🙏 

I mainly had some smaller comments, typo's etc. Feel free to merge once you're ready.

spacy_llm/models/langchain/model.py
```diff
@@ -0,0 +1,130 @@
...
+
+class LangChain:
+    def __init__(
+        self,
+        name: str,
+        api: str,
+        config: Dict[Any, Any],
+        query: Callable[
+            ["langchain.llms.base.BaseLLM", Iterable[_PromptType]],
+            Iterable[_ResponseType],
```
kabirkhan, line 23: This usage of `TypeVar`s doesn't really make sense. It's pretty much meaningless to do this unless the `TypeVar` is constrained and the query is generic. 

If `LangChain` as a class was generic and had some way to fill in these prompt/response types with e.g. a Pydantic model that was going to validate them then it would make more sense. 

But as this works now, they might as well be `Any`.

I think the same can be said for the `hf` models

spacy_llm/models/rest/openai/registry.py
rmitsch: See https://github.com/explosion/spacy-llm/pull/176#discussion_r1245418927.

spacy_llm/pipeline/llm.py
rmitsch: I'll specify `name` for clarity here. As for the default value - pro: less config, con: less explicit. I'm good with either. I'll leave it up to a follow-up PR to change that.

spacy_llm/models/rest/cohere/registry.py
rmitsch: It's a weird name, yes. We should document this explicitly (see [here](https://github.com/explosion/spacy-llm/pull/176#issuecomment-1611662901)).

migration_guide.md
```diff
@@ -0,0 +1,85 @@
...
+
+In `0.4.x`:
+```ini
+[components.llm.model]
+@llm_models = "langchain.OpenAI.v1"
+name = "gpt-3-5-turbo"
+config = {"temperature": 0.3}
+```
+
+</details>
```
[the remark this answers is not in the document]
rmitsch: We didn't document it, but I'll add a line.

spacy_llm/models/langchain/model.py
rmitsch: Fair point. I believe the `TypeVar`s were still in there from an earlier, different design. Removed in https://github.com/explosion/spacy-llm/pull/176/commits/57bc905f5cd505ce36f4fa3d89ab8f51092147e3.

usage_examples/README.md
rmitsch: Done.

PR #180, opened 2023-06-20
PromptNER based Chain-of-Thought prompting for span tasks

kabirkhan:
## Description

Implements Chain-of-Thought prompting for Span tasks

Based on https://arxiv.org/pdf/2305.15444.pdf

### Example Usage:

```python
import spacy

nlp = spacy.blank("en")
nlp.add_pipe(
    "llm",
    name="llm_ner",
    config={
        "task" : {
            "@llm_tasks": "spacy.NER.v3",
            "description": "Entities are the names of people and basketball teams. Names of sports are not entities. Names of soccer, baseball, rubgy, etc teams are not basketball team entities. adjectives are not entities. Pronouns are not persons. Man city is the name of a soccer team so not an entity in this case.",
            "examples": [
                {
                    "text": "The Golden State Warriors beat the LA Lakers by almost 20 points on the back of an amazing performance from Steph Curry.",
                    "entities": [
                         # In the final code these should be able to be provided in the prompt str format
                         # e.g. | Golden State Warriors | BASKETBALL_TEAM | True | is a basketball team in the NBA |
                        {
                            "text": "Golden State Warriors",
                            "label": "BASKETBALL_TEAM",
                            "is_entity": True,
                            "reason": "is a basketball team in the NBA"
                        },
                        {
                            "text": "LA Lakers",
                            "label": "BASKETBALL_TEAM",
                            "is_entity": True,
                            "reason": "is a basketball team in the NBA"
                        },
                        {
                            "text": "20",
                            "label": "==NONE==",
                            "is_entity": False,
                            "reason": "is a number"
                        },
                        {
                            "text": "amazing performance",
                            "label": "==NONE==",
                            "is_entity": False,
                            "reason": "is a description of an event"
                        },
                        {
                            "text": "Steph Curry",
                            "label": "PERSON",
                            "is_entity": True,
                            "reason": "is the name of a person"
                        },
                    ]
                }
            ],
            "labels": "PERSON,BASKETBALL_TEAM",
        },
        "model": {
            "@llm_models": "spacy.GPT-3-5.v1",
            "config": {"max_tokens": 200},
        },
        "save_io": True
    }
)

doc = nlp("The Denver Nuggets beat the Miami Heat 4 games to 1")
print([(ent.text, ent.label_) for ent in doc.ents])

# The names of these sports teams are specifically changed to view the model's reasoning behind labeling them or not.
doc2 = nlp("In other sports news Atletico Madrid was able to tie the Gonzaga bulldogs away from home with a stunning strike from Kevin De Bruyne.")
print([(ent.text, ent.label_) for ent in doc2.ents])
```

### Types of change

Feature

## Checklist

  - all tests requiring a GPU


spacy_llm/tasks/util/span.py
```diff
@@ -123,3 +124,116 @@ def _cfg_keys(self) -> List[str]:
     @property
     def _Example(self) -> Type[SpanExample]:
         return SpanExample
+
+
+class SpanTaskV2(SerializableTask[COTSpanExample]):
```
vin-ivar, line 133: Maybe it's a better idea to avoid defining a new class of `SpanTask` and just override the relevant methods with the new `NER` task instead?

vin-ivar: Hmm, actually I suppose it makes more sense this way, makes it more easily adaptable to other span tasks.

rmitsch: Yes, we prefer this to generalize to other span tasks.

spacy_llm/tasks/span.py
```diff
@@ -47,16 +85,33 @@ def __init__(
...
+                for example in self._prompt_examples
+                for key in example.entities
+            }
+        else:
+            example_labels = {
+                self._normalizer(key.label): key.label
+                for example in self._prompt_examples
+                for key in example.entities
+                if key.is_entity
+            }
```
rmitsch, line 113: The upper branch will be redundant once we only use CoT, right?

spacy_llm/tasks/util/examples.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class SpanReason(BaseModel):
+    text: str
+    is_entity: bool
+    label: str
+    reason: str
+
+    @classmethod
+    def from_str(cls, s: str, sep: str = "|"):
```
rmitsch, line 23: This would benefit from a docstring and some illustrative examples demonstrating the expected input.

spacy_llm/tasks/templates/ner.v3.jinja
```diff
@@ -0,0 +1,19 @@
...
+Q: Given the paragraph below, identify a list of possible entities, and for each entry explain why it is or is not an entity:
+{# whitespace #}
+{# whitespace #}
+{%- for example in examples -%}
+Paragraph: {{ example.text }}
+Answer:
+{# whitespace #}
+{%- for span in example.entities -%}
+{{ loop.index }}. {{ span.to_str() }}
+{# whitespace #}
```
rmitsch, line 13: This confuses me. The prompt says to do a binary yes/no classification, but here we only use positive examples, correct?

spacy_llm/tasks/util/examples.py
```diff
@@ -0,0 +1,44 @@
...
+    text: str
+    entities: Dict[str, List[str]]
+
+
+class LemmaExample(BaseModel):
+    text: str
+    lemmas: List[Dict[str, str]]
+
+
+class SpanReason(BaseModel):
```
rmitsch, line 16: This is a single reponse from the LLM, right?

rmitsch: Also, isn't this a duplicate of https://github.com/explosion/spacy-llm/pull/180/files#diff-86272b7a9396b646b75f0cf9f7bb4a10eaf34a0dd5cfc6d637d00be93fde050dR15?

spacy_llm/tasks/ner.py
```diff
@@ -203,10 +253,108 @@ def scorer(
...
+
+        self._label_dict = {self._normalizer(label): label for label in labels}
+
+    def _format_response(self, response: str) -> Iterable[Tuple[str, Iterable[str]]]:
+        """Parse raw string response into a structured format"""
+        output: dict[str, list[str]] = defaultdict(list)
+        assert self._normalizer is not None
+        for line in response.strip().split("\n"):
+            entity = SpanReason.from_str(line)
+            if entity:
```
rmitsch, line 336: 1. I'd rename `entity` to `span_reason` or similar.
2. `SpanReason.from_str()` only returns a `SpanReason` instance if `line` was parsed successfully for an entity, right?

rmitsch (commented):
We're getting there :slightly_smiling_face: The templates for the other span-based tasks need to be updated still.

spacy_llm/tasks/util/examples.py
vin-ivar: Oof, fixed!

spacy_llm/tasks/ner.py
vin-ivar: Yeah, it does

spacy_llm/tasks/util/examples.py
```diff
@@ -0,0 +1,8 @@
+from typing import Dict, List
```
kabirkhan, line 1: You could delete this file now since this is inlined in the `lemma.py` task

spacy_llm/ty.py
```diff
@@ -17,7 +17,7 @@
 
 PromptExecutor = Callable[[Iterable[_Prompt]], Iterable[_Response]]
 ExamplesConfigType = Union[
```
kabirkhan, line 19: This can still be None for other tasks (and actually I think we could maybe still allow it to None for this COT version of the NER task)

spacy_llm/tasks/templates/ner.v3.jinja
```diff
@@ -0,0 +1,19 @@
+{{ description }}
```
kabirkhan, line 1: We should also support `label_definitions` here similar to the older prompts. That could be a more structured way to provide the description.

vin-ivar: Yeah I'm wondering what the best way to do this is. The original PromptNER paper has something that resembles label definitions - each entity is kind of fleshed out in the first sentence of the prompt, and the labels are defined in the second sentence, so maybe I could just deterministically generate something similar?

spacy_llm/tasks/ner.py
```diff
@@ -11,69 +12,31 @@
...
-    examples: ExamplesConfigType = None,
     normalizer: Optional[Callable[[str], str]] = None,
     alignment_mode: Literal["strict", "contract", "expand"] = "contract",
     case_sensitive_matching: bool = False,
     single_match: bool = False,
 ):
-    """NER.v2 task factory.
+    """NER.v3 task factory.
 
+    examples (Union[Callable[[], Iterable[OTS]]]): Optional callable that
```
rmitsch, line 37: Where's `OTS` coming from?

spacy_llm/tasks/span.py
```diff
@@ -47,16 +78,37 @@ def __init__(
...
+        null_labels = {
+            self._normalizer(entity.label): entity.label
+            for example in self._prompt_examples
+            for entity in example.entities
+            if not entity.is_entity
+        }
+        if len(null_labels) > 1:
+            warnings.warn(
+                f"Negative examples contain multiple negative labels: {', '.join(null_labels.keys())}."
+            )
```
rmitsch, line 103: I'm not sure I understand this check. Could you elaborate on why this is necessary?

spacy_llm/tasks/span.py
```diff
@@ -143,6 +172,20 @@ def assign_spans(
         """Assign spans to the document."""
         raise NotImplementedError()
 
+    def _format_response(self, response: str) -> Iterable[Tuple[str, Iterable[str]]]:
+        """Parse raw string response into a structured format"""
+        output: dict[str, list[str]] = defaultdict(list)
```
rmitsch, line 177: Let's do `Dict` and `List` so we don't have to do `from __future__ import annotation` (we support Python < 3.9 too).

spacy_llm/tasks/templates/ner.v3.jinja
rmitsch: I agree, would be great to make this more modular.

spacy_llm/tasks/span.py
vin-ivar: You have to have some kind of label for negative examples, we just use a random string like `none`. It might be a bit odd if someone were to use multiple different variants, so it probably makes sense to inform them.

spacy_llm/tasks/ner.py
vin-ivar: Not sure, this wasn't me - @kabirkhan?

spacy_llm/tests/tasks/test_ner.py
```diff
@@ -323,7 +323,7 @@ def test_ensure_offsets_correspond_to_substrings(
 )
 def test_ner_zero_shot_task(text, response, gold_ents):
     labels = "PER,ORG,LOC"
-    llm_ner = make_ner_task_v2(labels=labels)
+    llm_ner = make_ner_task_v3(labels=labels)
```
svlandeg, line 326: I'm surprised that only the task factory is changed in these tests, and the test themselves have remained unchanged. Doesn't the `response` look quite different with the v3 task? Where are we testing that the new format is actually parsed correctly on a realistic end-to-end example?

spacy_llm/tasks/templates/ner.v3.jinja
kabirkhan: I would just put the `label_definitions` blurb from the `ner.v2` template right below description. 

Users could provide definitions inline in the description (like how I did it in the example of the PR body) or they could provide the dict of `LABEL: definition` with the description being a simple 1 line description of the task

spacy_llm/tasks/ner.py
kabirkhan: Wasn't me either I don't think haha. This should be Just be `ExamplesConfigType`

spacy_llm/tests/tasks/test_spancat.py
```diff
@@ -318,43 +390,43 @@ def test_spancat_alignment(response, alignment_mode, gold_spans):
...
-            [("Jean", "PER"), ("Jean", "PER")],
-        ),
-        (
-            "PER: Jean",
-            True,
+            "1. Jean | True | PER | is a person's name",
             True,
             [("Jean", "PER")],
         ),
+        # TODO: Get this matching working again, issue is with the parsing
```
kabirkhan, line 409: This TestCase doesn't work and it's pretty tough to get it to work properly:

spacy_llm/tasks/span.py
```diff
@@ -143,39 +196,76 @@ def assign_spans(
...
+            label = self._label_dict[norm_label]
+            span_reason.label = label
+            span_reasons.append(span_reason)
+        return span_reasons
+
     def parse_responses(
         self, docs: Iterable[Doc], responses: Iterable[str]
     ) -> Iterable[Doc]:
-        for doc, prompt_response in zip(docs, responses):
+        """Since we provide entities in a numbered list, we expect the LLM to
```
kabirkhan, line 220: The way we parse the responses needs to change entirely. 

We now assume the entities are reported from the LLM in order and we map entities/spans incrementally based utilizing the `find_after` param of `find_substrings`. 

So for input text: 

```
"The Denver Nuggets beat the Miami Heat 4 games to 1"
```

and a response:

```
1. Denver Nuggets | True | BASKETBALL_TEAM | is a basketball team in the NBA
2. Miami Heat | True | BASKETBALL_TEAM | is a basketball team in the NBA
3. 4 | False | ==NONE== | is a number
4. games | False | ==NONE== | is a common noun
5. 1 | False | ==NONE== | is a number
```


we first parse the list of `SpanReason`s which is easy, break by newline and call `SpanReason.from_str(line)`.

When we map back to spaCy spans now, we have some ordered structured so we no longer need to do `single_match=False` (and in fact that setting leads to false positives so it's good we're removing it)

So for ent 1, we match against the whole input text.

1. Matching `Denver Nuggets`

Text to match against
```
"The [Denver Nuggets](Found) beat the Miami Heat 4 games to 1"
```

Now we consider only the text after this:

For NER, we find after the `end_char` of the last found span, for SpanCat, we consider overlapping entities so we use the `start_char`.

2. Matching "Miami Heat"

Text to match against:
```
" beat the [Miami Heat](FOUND) 4 games to 1"
```


This incremental parsing of the span reasons against a smaller and smaller section of the input string works great most of the time. Issues arise with this specific test case and Spancat.
https://github.com/explosion/spacy-llm/pull/180/files#r1279693295

kabirkhan: The big open question here now is should we prioritize `case_sensitive` matches? I think we should. I think if the case is correct in the LLM response and we can find a match let's use that. Then fallback to `case_insensitive` only if no case sensitive matches are found for a line of output.

So for input text:

```
This guy jean (or Jean) is the president of the Jean Foundation.
```

And output:

```
1. Jean | True | PER | is a person's name
```

with `case_sensitive=False`, we would have marked:

```
This guy [jean](PERSON) (or Jean) is the president of the Jean Foundation.
```

But I think first we should scan for case sensitive matches, then fallback afterward.

So we'd get a final output of:

```
This guy jean (or [Jean](PERSON)) is the president of the Jean Foundation.
```

regardless of whether `case_sensitive` is True or False.

kabirkhan: @svlandeg @rmitsch if you could weigh in here that would be awesome!

rmitsch: - Incremental parsing seems reasonable.
- Same for prioritizing case-sensitive matching.
- Commenting on the test case further above.

spacy_llm/tests/tasks/test_spancat.py
rmitsch: Can you elaborate on why it's tough?

kabirkhan: This is the test case referenced below

spacy_llm/tasks/span.py
kabirkhan: Sweet, I'll keep these changes as is then :)

spacy_llm/tests/tasks/test_spancat.py
rmitsch: `This guy jean (or Jean) is the president of the Jean Foundation.`, right? Is it still difficult to make this work with the different parsing you proposed?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,6 +7,9 @@ ran. If your test fixes a bug reported in an issue, don't forget to include the
 issue number. If your PR is still a work in progress, that's totally fine – just
 include a note to let us know. -->
 
+### Corresponding documentation PR
+<!--- Add the link to the corresponding documentation PR here, if applicable. -->
+
 ### Types of change
```
svlandeg, line 13: These edits shouldn't be showing up in the diff of this PR. Can you merge in the current `develop` so they hopefully go away? It'll allow us to focus review on the actual differences this PR introduces.

spacy_llm/tasks/span.py
svlandeg: Hm, I'm not sure I agree. If `case_sensitive` is set specifically to `False`, then there shouldn't be some hidden logic that still prioritizes case-matching variants.

svlandeg: I do think that in general the way results are parsed now seems much better - i.e. requiring an order and needing two SpanReasons rather than relying on a "match all" or "match once" logic. 

On our internal board, we had a similar example that we wanted to fix, but that we had blocked on this PR because I assumed this PR would solve it:

In the sentence "The woman Paris was walking around in Paris, talking to her friend Paris" we'd need the LLM to return something like

Paris: PER
Paris: LOC
Paris: PER

in order, to be able to parse this example correctly.

spacy_llm/tests/tasks/test_ner.py
```diff
@@ -868,15 +764,112 @@ def test_label_inconsistency():
...
+
+@pytest.mark.parametrize(
+    "text,person_first",
+    [
+        ("The woman Paris was walking around in Paris", True),
+        ("Walking around Paris as a woman named Paris is quite a trip.", False),
+    ],
+    ids=["person_first", "location_first"],
+)
+def test_regression_span_task_response_parse(text: str, person_first: bool):
```
kabirkhan, line 821: Regression test

spacy_llm/tasks/span.py
kabirkhan: Hmm ok I see your point, I could remove the separate hidden case sensitive match step. That's probably for the best. As for order of extraction, I saw that issue in GitHub already and added a specific regression test for this sentence: https://github.com/explosion/spacy-llm/pull/180/files#r1282197256

svlandeg: @kabirkhan : the regression test has only two instances though, I wanted to test specifically with two PERSON entities and a LOC inbetween. Can the current implementation deal with that?

kabirkhan: Ok well actually, the issue is this sentence referenced above doesn't work without the case sensitive matching step and the way I'm using the `find_after` param with spancat.

Input:
```
This guy jean (or Jean) is the president of the Jean Foundation.
```

Response:

```
1. jean | True | PER | is a person's name
2. Jean | True | PER | is a person's name
3. Jean Foundation | True | ORG | is the name of an Organization name
```


### Approach without just case insensitive matching

* Will find the first "jean" and mark the `find_after` parameter to the `start_char` of "jean" because we want to allow overlapping spans.
* Will then look for "Jean" but case insensitive ("jean") and since the parsing starts at the `start_char` (not the `end_char` of the first "jean", we mark that span again, then move on to the next entity "Jean Foundation"
* "Jean Foundation" gets parsed correctly

Final spans list is: `["jean", "jean", "Jean Foundation"]

We end up doing a `set(spans)` to remove identical spans so we only get 2 spans returned from this response of 3 spans.

This is only for Spancat with `allow_overlap=True`. NER works correctly since we set the `find_after` to the end of the previously found ent. If we did this for `Spancat` we'd never parse out overlapping spans.

spacy_llm/tests/tasks/test_spancat.py
kabirkhan: Yes it is, unless we do the separate step of first case sensitive matching, then case insensitive matching afterward. Explained in more detail below

spacy_llm/tasks/span.py
kabirkhan: > @kabirkhan : the regression test has only two instances though, I wanted to test specifically with two PERSON entities and a LOC inbetween. Can the current implementation deal with that?

yep! as long as the LLM responds with all 3 entities the parsing will work for NER or for SpanCat with allow_overlap=False. If allow_overlap is True for Spancat, the current implementation won't work. 

Also I fixed the issue with the "jean", "Jean", "Jean Foundation" parsing in the latest commit.

kabirkhan: The only way to get that sentence to work with `SpanCat` and `allow_overlap=True` would be to alter the prompt in some way that LLM marks a position in the text when it extracts SpanReason's. Having the LLM respond with `start_char` and `end_char` didn't really work at all when I tried. 

I definitely think it's something we could try to address in another new version of the model with something like this Paper's implementation: https://arxiv.org/abs/2304.10428 where they mark offsets in the original text

svlandeg (commented):
Few quick comments, more in-depth review in process ;-)

spacy_llm/tasks/span.py
rmitsch: > I definitely think it's something we could try to address in another new version of the model with something like this Paper's implementation: https://arxiv.org/abs/2304.10428 where they mark offsets in the original text

I use this kind of in-text markers for the EL task, and it seems to work quite well (still have to run experiments to get numbers).

> The only way to get that sentence to work with SpanCat and allow_overlap=True would be to alter the prompt in some way that LLM marks a position in the text when it extracts SpanReason's. Having the LLM respond with start_char and end_char didn't really work at all when I tried.

I would have assumed that with `SpanCat` `allow_overlap is True` always applies? Anyway, if that's a limitation of the current approach, that's ok. We can fix that/use in-text markers in a later PR, but let's wrap this up.

spacy_llm/tasks/spancat.py
```diff
@@ -1,78 +1,37 @@
...
-from ..ty import ExamplesConfigType
+from ..ty import RequiredExamplesConfigType
 from ..util import split_labels
-from .span import SpanExample, SpanTask
+from .span import SpanExample, SpanReason, SpanTask
 from .templates import read_template
 
-_DEFAULT_SPANCAT_TEMPLATE_V1 = read_template("spancat")
-_DEFAULT_SPANCAT_TEMPLATE_V2 = read_template("spancat.v2")
+_DEFAULT_SPANCAT_TEMPLATE_V3 = read_template("spans.v3")
```
svlandeg, line 16: Why not `spancat.v3` to be in line with previous files?

usage_examples/ner_v3_openai/README.md
```diff
@@ -0,0 +1,77 @@
+# Using GPT Models from OpenAI for Named Entity Recognition (NER)
+
+
+This example shows how you can use a model from OpenAI for Named Entity Recognition (NER).
+The NER prompt is based on the the [PromptNER](https://arxiv.org/abs/2305.15444) paper and
```
svlandeg, line 5:
```suggestion
The NER prompt is based on the [PromptNER](https://arxiv.org/abs/2305.15444) paper and
```

usage_examples/ner_v3_openai/README.md
```diff
@@ -0,0 +1,77 @@
...
+utilizes Chain-of-Thought reasoning to extract named entities. It requires at least 1 example
+and does not work with zeroshot settings.
```
svlandeg, line 7: We should guard against this specifically: when given an empty list as examples, the config will parse and zero scores will come out.

usage_examples/tests/test_readme_examples.py
```diff
@@ -54,8 +55,8 @@ def test_example_2_ner_hf():
         factory = "llm"
 
         [components.llm.task]
-        @llm_tasks = "spacy.NER.v2"
-        labels = ["PERSON", "ORGANISATION", "LOCATION"]
+        @llm_tasks = "spacy.TextCat.v2"
+        labels = ["COMPLIMENT", "INSULT"]
```
svlandeg, line 60: This could be an additional test, but I don't see why we're removing the old one for `spacy.NER.v2`? As we discussed that we wouldn't actually be deprecating these old versions, let's keep their tests intact as well.

requirements-dev.txt
```diff
@@ -7,7 +7,7 @@ mypy>=0.990,<1.1.0; platform_machine != "aarch64" and python_version >= "3.7"
 black==22.3.0
 types-requests==2.28.11.16
 # Prompting libraries needed for testing
-langchain==0.0.219; python_version>="3.9"
+langchain==0.0.249; python_version>="3.9"
```
svlandeg, line 10: There are still unrelated changes showing up in the diff of this PR, can that be fixed @kabirkhan ?

svlandeg: This change is already present in both `main` and `develop`.

spacy_llm/tasks/ner.py
```diff
@@ -203,10 +134,30 @@ def scorer(
...
+        return [
+            "_label_dict",
+            "_template",
+            "_description",
+            "_label_definitions",
+            "_alignment_mode",
+            "_case_sensitive_matching",
+        ]
+
     def _create_prompt_example(self, example: Example) -> SpanExample:
```
kabirkhan, line 148: Ref for creating prompt examples from examples.

spacy_llm/tasks/spancat.py
kabirkhan: The spancat and ner templates were duplicated except the 1 line allowing overlapping spans. So I consolidated them into a single file. I could revert if we want to keep these standard but it's an internal implementation detail so we could always change later if the templates need to be more different than they are now.

spacy_llm/tasks/ner.py
```diff
@@ -168,26 +104,21 @@ def initialize(
...
+        for eg in examples:
             if n_prompt_examples < 0 or len(self._prompt_examples) < n_prompt_examples:
                 self._prompt_examples.append(self._create_prompt_example(eg))
-
-        self._label_dict = {
-            self._normalizer(label): label for label in sorted(set(labels))
-        }
+        if not labels:
+            label_set = set()
+            for eg in examples:
```
svlandeg, line 117: These `for` loops got rewritten, but they were using `infer_labels` on purpose so that we don't have to store all the `get_examples()` in a variable and we can actually just loop through them once instead of twice. I think the original implementation was more efficient.

spacy_llm/tasks/ner.py
```diff
@@ -168,26 +104,21 @@ def initialize(
...
+                for ent in eg.reference.ents:
+                    label_set.add(ent.label_)
+            labels = sorted(set(label_set))
```
svlandeg, line 120: In the reimplementation, `label_set` is already a `set`, so no need to do `set()` again.

usage_examples/tests/test_readme_examples.py
rmitsch: Tests for old versions of NER and SpanCat are in `tests/legacy` (not sure whether that's the ideal location, but good enough for now).

spacy_llm/tasks/span.py
rmitsch: I think we can close this?

svlandeg: Do we now have a unit test for the original example I mentioned (with the three entities, first PER, then LOC, then PER again)?

spacy_llm/tasks/ner.py
```diff
@@ -9,71 +8,30 @@
...
-        case_sensitive_matching=case_sensitive_matching,
-        single_match=single_match,
-    )
-
-
-@registry.llm_tasks("spacy.NER.v2")
-def make_ner_task_v2(
+@registry.llm_tasks("spacy.NER.v3")
+def make_ner_task_v3(
+    examples: RequiredExamplesConfigType,
```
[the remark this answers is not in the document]
rmitsch: Can be an empty list now. If so, the template renders the default example.

usage_examples/ner_v3_openai/README.md
rmitsch: Removed "It requires at least 1 example and does not work with zeroshot settings.", as we have a default example now.

usage_examples/textcat_dolly/README.md
```diff
@@ -1,9 +1,9 @@
 # Using open-source Dolly models hosted on Huggingface
 
 This example shows how you can use the [open-source Dolly
-models](https://github.com/databrickslabs/dolly) hosted on Huggingface in a
-Named Entity Recognition (NER) task. We demonstrate how you can use large
-language models in zero- or few-shot annotation settings. 
+models](https://github.com/databrickslabs/dolly) hosted on Huggingface for categorizing texts in
+zero- or few-shot settings. Here, we perform binary text classification to
+determine if a given text is an `INSULT` or a `COMPLIMENT`.
```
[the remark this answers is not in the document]
rmitsch: Did you comment on the wrong file? Which directories wouldn't you want renamed?

spacy_llm/tasks/ner.py
rmitsch: Fixed with refactoring.

usage_examples/textcat_dolly/README.md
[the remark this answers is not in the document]
svlandeg: We can have "usage_examples/textcat_dolly" as an additional example, but we should keep ""usage_examples/ner_dolly" as is, even if that means it uses v2. 

(personally I would move all of this out of this PR for easier reviewing & discussion)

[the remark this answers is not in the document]
svlandeg: Better yet, we should try whether this example (the original NER Dolly one) works with NER.v3.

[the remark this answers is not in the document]
rmitsch: Readded `usage_examples/ner_dolly`, removed `usage_examples/textcat_dolly`. The latter is [here](https://github.com/explosion/spacy-llm/pull/271).

rmitsch:
I think this is ready for a final once-over and running the external tests in the CI.

svlandeg (commented):
There's still two unresolved conversations, can you have a look at those?

I'd also like us to run a few more tests on how v3 can be used as a substitute for v2, and which changes (if any) would be needed in e.g. the Dolly NER example.

spacy_llm/tests/tasks/test_ner.py
rmitsch: > I'm surprised that only the task factory is changed in these tests, and the test themselves have remained unchanged. 

As long as the examples have been adjusted to include span reasons, this should work.

> Where are we testing that the new format is actually parsed correctly on a realistic end-to-end example?

What do you consider realistic? This test includes a few IMO realistic sentences. We could be more thorough, but this is sufficient in my opinion.

svlandeg: You're right that in terms of end-to-end example this should be sufficient. I think I was confused also by the fact that we have tests that parse the actual prompt to check the text is formatted as we expect. I don't see any of those tests added for v3?

spacy_llm/tasks/span.py
rmitsch: No, I'll add one.

rmitsch: See https://github.com/explosion/spacy-llm/pull/180/commits/2979ffc9f5e077ffc5f84619ad918ddabe7c95d0. Had to make some adjustments to the config to make the example work.

svlandeg: Awesome, thanks!

.github/workflows/test.yml
```diff
@@ -19,7 +19,7 @@ env:
 jobs:
   run:
     strategy:
-      fail-fast: true
+      fail-fast: false
```
svlandeg, line 22:
```suggestion
      fail-fast: true
```
Reverting this back before merging, right?

spacy_llm/tasks/ner/registry.py
```diff
@@ -111,4 +116,66 @@ def make_ner_task_v2(
...
+):
+    """NER.v3 task factory, with chain-of-thought prompting.
+
+    parse_responses (Optional[TaskResponseParser[SpanTask]]): Callable for parsing LLM responses for this task.
+    prompt_example_type (Optional[Type[FewshotExample]]): Type to use for fewshot examples.
+    labels (Union[str, List[str]]): List of labels to pass to the template,
+        either an actual list or a comma-separated string.
+        Leave empty to populate it at initialization time (only if examples are provided).
+    template (str): Prompt template passed to the model.
+    description (str): todo
```
svlandeg, line 146: TODO

spacy_llm/tasks/ner/registry.py
```diff
@@ -111,4 +116,66 @@ def make_ner_task_v2(
...
+        of the label to help the language model output the entities wanted.
+        It is usually easier to provide these definitions rather than
+        full examples, although both can be provided.
+    examples (Optional[Callable[[], Iterable[Any]]]): Optional callable that reads a file containing task examples for
+        few-shot learning. If None is passed, then zero-shot learning will be used.
+    normalizer (Optional[Callable[[str], str]]): optional normalizer function.
+    alignment_mode (str): "strict", "contract" or "expand".
+    case_sensitive_matching (bool): Whether to search without case sensitivity.
+    single_match (bool): If False, allow one substring to match multiple times in
+        the text. If True, returns the first hit.
```
svlandeg, line 157:
```suggestion
```

spacy_llm/tasks/ner/task.py
```diff
@@ -47,6 +51,8 @@ def __init__(
         single_match (bool): If False, allow one substring to match multiple times in
             the text. If True, returns the first hit.
         scorer (Scorer): Scorer function.
+        description (str): todo
```
svlandeg, line 54: TODO

spacy_llm/tasks/spancat/registry.py
```diff
@@ -119,4 +126,72 @@ def make_spancat_task_v2(
...
+        of the label to help the language model output the entities wanted.
+        It is usually easier to provide these definitions rather than
+        full examples, although both can be provided.
+    examples (Optional[Callable[[], Iterable[Any]]]): Optional callable that reads a file containing task examples for
+        few-shot learning. If None is passed, then zero-shot learning will be used.
+    normalizer (Optional[Callable[[str], str]]): optional normalizer function.
+    alignment_mode (str): "strict", "contract" or "expand".
+    case_sensitive_matching (bool): Whether to search without case sensitivity.
+    single_match (bool): If False, allow one substring to match multiple times in
+        the text. If True, returns the first hit.
```
svlandeg, line 170:
```suggestion
```

spacy_llm/tasks/spancat/registry.py
```diff
@@ -119,4 +126,72 @@ def make_spancat_task_v2(
...
+    """SpanCat.v3 task factory for SpanCat with chain-of-thought prompting.
+
+    parse_responses (Optional[TaskResponseParser[SpanCatTask]]): Callable for parsing LLM responses for this
+        task.
+    prompt_example_type (Optional[Type[FewshotExample]]): Type to use for fewshot examples.
+    labels (Union[str, List[str]]): List of labels to pass to the template,
+        either an actual list or a comma-separated string.
+        Leave empty to populate it at initialization time (only if examples are provided).
+    template (str): Prompt template passed to the model.
+    description (str): todo
```
svlandeg, line 159: TODO

spacy_llm/tasks/templates/ner.v3.jinja
```diff
@@ -0,0 +1,64 @@
+You are an expert Named Entity Recognition (NER) system.
+Your task is to accept Text as input and extract named entities.
+{%- if allow_overlap -%}
+{# whitespace #}
+The entities you extract can overlap with each other.
```
svlandeg, line 5: Wait, should we have this in the NER template? Shouldn't we ask for non-overlapping spans, and only allow overlap for the spancat template?

spacy_llm/tasks/templates/spancat.v3.jinja
```diff
@@ -0,0 +1,64 @@
+You are an expert Named Entity Recognition (NER) system.
```
svlandeg, line 1: For spancat, should we call this differently, maybe at least drop the "named" bit? I'm not sure though, maybe NER + overlapping allowed is fine.

spacy_llm/tests/tasks/test_ner.py
```diff
@@ -209,47 +221,61 @@ def test_ner_config(cfg_string, request):
...
-        "zeroshot_cfg_string_v2_lds",
-        "fewshot_cfg_string",
-        "fewshot_cfg_string_v2",
-        "ext_template_cfg_string",
+        # simple
+        # (
+        #     "Marc and Bob both live in Ireland.",
+        #     [("Marc", "PER"), ("Bob", "PER"), ("Ireland", "LOC")],
+        # ),
+        # same entity names
```
svlandeg, line 235:
```suggestion
```

spacy_llm/tests/tasks/test_ner.py
```diff
@@ -868,15 +864,88 @@ def test_label_inconsistency():
...
+            "2. Paris | True | PER | is the name of the woman\n",
+            [("Paris", "LOC"), ("Paris", "PER")],
+        ),
+    ],
+    ids=["3_ents", "2_ents"],
+)
+def test_regression_span_task_response_parse(
+    text: str, response: str, gold_ents: List[Tuple[str, str]]
+):
+    """Test based on on spaCy issue: https://github.com/explosion/spaCy/discussions/12845
```
svlandeg, line 935:
```suggestion
    """Test based on spaCy issue: https://github.com/explosion/spaCy/discussions/12812
```
(which is the original issue where this example came from)

svlandeg (commented):
This is looking really good, nice team effort all! 🙏 

Had a few nits + one more important question about a phrasing in the NER template.

spacy_llm/tasks/templates/ner.v3.jinja
rmitsch: It was originally one template for both SpanCat and NER. For consistency reasons I split it in two and named them after the task versions. 
Both templates are fixed in https://github.com/explosion/spacy-llm/pull/180/commits/c763689fbd95c02a7d64b13d33a45a89c1edc14e.

spacy_llm/tasks/templates/spancat.v3.jinja
rmitsch: Done in https://github.com/explosion/spacy-llm/pull/180/commits/1c269e04afea70397dbfcf558f4355b6b6894e24.

spacy_llm/tasks/spancat/registry.py
rmitsch: Done in https://github.com/explosion/spacy-llm/pull/180/commits/3a31530bfd257cd77028a964e9b1e421676c0abd.

spacy_llm/tasks/ner/task.py
rmitsch: Done in https://github.com/explosion/spacy-llm/pull/180/commits/3a31530bfd257cd77028a964e9b1e421676c0abd.

spacy_llm/tasks/ner/registry.py
rmitsch: Done in https://github.com/explosion/spacy-llm/pull/180/commits/3a31530bfd257cd77028a964e9b1e421676c0abd.

PR #183, opened 2023-06-23
Detect label inconsistency in `SpanTask`

rmitsch:
## Description

Check for inconsistency between specified labels and labels in examples.

## Checklist

  - all tests requiring a GPU


spacy_llm/tasks/util/span.py
```diff
@@ -36,6 +36,27 @@ def __init__(
...
+                for example in self._examples
+                for key in example.entities
+            }
+        )
+        spec_labels = list(self._label_dict.keys())
+        example_labels.sort()
+        spec_labels.sort()
+        if not set(example_labels) <= set(spec_labels):
+            raise ValueError(
+                f"Specified labels and labels in examples diverge. Received {spec_labels} and {example_labels}, correspondingly. Please ensure your label specification and example labels are consistent."
```
svlandeg, line 57: I would make this warning message more specific and say 
1) which label set was defined as part of the task definition 
2) which labels were found in the examples that were NOT part of the task definition.

We can compile list 2) by keeping track of the labels not in `spec_labels`.

This will ultimately help users understand more quickly what is going on and which parts of their configuration they need to adjust.

I think this approach will also allow us to output the actual original labels instead of the normalized ones. We'd store the original in the deviating list if it's normalized form is not in `spec_labels`.

spacy_llm/tests/tasks/examples/ner_inconsistent.yml
```diff
@@ -0,0 +1,11 @@
+- text: Jack and Jill went up the hill.
+  entities:
+    PERSON:
+      - Jack
+      - Jill
+    location:
```
svlandeg, line 6:
```suggestion
    LOCATION:
```
(it doesn't really matter, but just for consistency)

svlandeg (commented):
We definitely need to think about how to handle such an inconsistency.

I think we should warn instead of error here though, and discard the examples that do not fall into the label list that was specified earlier. Basically that allows users to keep the list of examples fixed while playing with different labels for the LLM to recognize (sometimes adding/removing labels can influence accuracy within the same label category).

spacy_llm/tasks/util/span.py
rmitsch: Like in https://github.com/explosion/spacy-llm/pull/183/commits/39606aaefbea9950369f4d2a6ea2a65a21c1021c?

rmitsch:
> Basically I think we can recover by just keeping to the pre-defined list of labels, and discard the non-matching ones, while warning?

Done in https://github.com/explosion/spacy-llm/pull/183/commits/d2aadb3658569075fea4d010c557f53a554e87ac.

svlandeg (approved):
Virtually good to merge, just two smaller comments

spacy_llm/tasks/util/span.py
```diff
@@ -36,6 +37,48 @@ def __init__(
...
         self._single_match = single_match
 
+        if self._examples:
+            self._examples = self._check_label_consistency()
+
+    def _check_label_consistency(self) -> List[SpanExample]:
+        """Checks consistency of labels between examples and defined labels. Emits warning on inconsistency.
+        RETURNS ():
+        """
+        assert self._examples
```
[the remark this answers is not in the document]
rmitsch: Right, this is just because `mypy` complains otherwise :)

spacy_llm/tasks/util/span.py
```diff
@@ -36,6 +37,48 @@ def __init__(
         self._case_sensitive_matching = case_sensitive_matching
         self._single_match = single_match
 
+        if self._examples:
+            self._examples = self._check_label_consistency()
+
+    def _check_label_consistency(self) -> List[SpanExample]:
+        """Checks consistency of labels between examples and defined labels. Emits warning on inconsistency.
+        RETURNS ():
```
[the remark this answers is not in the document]
rmitsch: Done in https://github.com/explosion/spacy-llm/pull/183/commits/fc2f901551fd9ef4feda6aeca068aba3f96821b3.

PR #188, opened 2023-06-24
Fix cache and nlp.pipe(..., as_tuples=True)

kabirkhan:
## Description

Cached docs currently don't pass along the `doc._context` attr that `nlp.pipe` is called with when using `as_tuples=True`. This ensures that even cached docs have the correct context.

### Types of change

Fix

## Checklist

  - all tests requiring a GPU


rmitsch (approved):
Thanks!

adrianeboyd:
I'll say I'm not super-keen on this use of `doc._context`, which is internals and may change in the future.

rmitsch:
@adrianeboyd Do you see a better way to solve this problem?

PR #189, opened 2023-06-26
Update README.md

koaning:
With the comma in place there ... the command causes this error to appear: 

```
catalogue.RegistryError: [E893] Could not find function '"spacy.BatchCache.v1",' 
in function registry 'llm_misc'. If you're using a custom function, make sure the 
code is available. If the function is provided by a third-party package, e.g. 
spacy-transformers, make sure the package is installed in your environment.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1000,7 +1000,7 @@ Example config block:
 
 ```ini
 [components.llm.cache]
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Thanks!

PR #192, opened 2023-06-26
Use prompt template to ensure data consistency in cache directory

rmitsch:
## Description

Use prompt template to ensure data consistency in cache directory.

## Checklist

  - all tests requiring a GPU


spacy_llm/ty.py
```diff
@@ -104,6 +108,18 @@ def add(self, doc: Doc) -> None:
         doc (Doc): Doc to add to persistence queue.
         """
 
+    @property
+    def prompt_template(self) -> Optional[str]:
+        """Set prompt template.
```
shadeMe, line 113:
```suggestion
        """Get prompt template.
```

spacy_llm/ty.py
```diff
@@ -104,6 +108,18 @@ def add(self, doc: Doc) -> None:
...
+        RETURNS (str): Prompt template string used for docs to cache/cached docs.
```
shadeMe, line 114:
```suggestion
        RETURNS (Optional[str]): Prompt template string used for docs to cache/cached docs.
```

spacy_llm/pipeline/llm.py
```diff
@@ -117,9 +117,10 @@ def __init__(
         self._name = name
         self._task = task
         self._backend = backend
+        self._save_io = save_io
         self._cache = cache
         self._cache.vocab = vocab
-        self._save_io = save_io
+        self._cache.prompt_template = self._task.prompt_template
```
kabirkhan, line 123: Hmm I don't really like this setup where we override some attributes of the cache but also allow it to be provided on init. This sort of hidden override is hard to reason about. I missed it when we first did this for the vocab.

spacy_llm/cache.py
```diff
@@ -62,7 +65,52 @@ def __init__(
...
+                existing_prompt_template
+            ) != BatchCache._normalize_prompt_template(self._prompt_template):
+                raise ValueError(
+                    f"Prompt template in cache directory ({prompt_template_path}) is not equal with "
+                    f"current prompt template. Reset your cache if you are using a new prompt "
+                    f"template."
+                )
+
+    @staticmethod
+    def _normalize_prompt_template(prompt_template: str) -> str:
```
kabirkhan, line 108: Why not just hash the prompt template here rather than doing this normalization? Whitespace/newlines also matter for what the LLM returns sometimes so if you add some newlines it could easily change the response.

spacy_llm/pipeline/llm.py
rmitsch: The issue here is that we don't have this information (vocab and prompt template) on cache init time. I'm not a fan of doing it this way either, but don't see another way to do this.

> Hmm I don't really like this setup where we override some attributes of the cache but also allow it to be provided on init.

Neither `prompt_template` nor `vocab` can be provided on init.

spacy_llm/cache.py
rmitsch: Hm, ok. I assumed that whitespace doesn't matter. In that case I didn't want to enforce rebuilding the entire cache just because users added a newline for nicer formatting.

adrianeboyd: Would it make sense to have the prompt be the cache key?

rmitsch: I think not, but depends on what you mean specifically. We don't want docs modified via different prompt templates. "Cache key" as I understand it in this context is the hash assigned to a batch/doc bin, which doesn't need the prompt template involved as there should be only one prompt template per cache directory.

rmitsch: Updated to hash in https://github.com/explosion/spacy-llm/pull/192/commits/31edb7b9abd47726198523515123cc3151e47fc9.

spacy_llm/pipeline/llm.py
kabirkhan: Ok I see what you mean here. Could we maybe add an `initialize` method like:

```
def initialize(self, vocab: Vocab, task: LLMTask) -> None:
     ...
```

It's kind of a nit but this code path really seems like it could easily lead to bugs down the road

rmitsch: Like in https://github.com/explosion/spacy-llm/pull/192/commits/62ec06bd234ead286610fd84293fadbd6f58e821?

kabirkhan: Exactly yeah, looks good to me

README.md
```diff
@@ -1388,23 +1388,27 @@ The default `query` (`spacy.CallLangChain.v1`) executes the prompts by running `
...
-| `path`               | `Optional[Union[str, Path]]` | `None`  | Cache directory. If `None`, no caching is performed. |
-| `batch_size`         | `int`                        | 64      | Number of docs in one batch (file).                  |
-| `max_batches_in_mem` | `int`                        | 4       | Max. number of batches to hold in memory.            |
+| Argument             | Type                         | Default | Description                                                                                                               |
+| -------------------- | ---------------------------- | ------- | ------------------------------------------------------------------------------------------------------------------------- |
+| `path`               | `Optional[Union[str, Path]]` | `None`  | Cache directory. If `None`, no caching is performed, and this component will act as a NoOp.                               |
+| `batch_size`         | `int`                        | 64      | Number of docs in one batch (file). Once a batch is full, it will be peristed to disk.                                    |
```
svlandeg, line 1407:
```suggestion
| `batch_size`         | `int`                        | 64      | Number of docs in one batch (file). Once a batch is full, it will be persisted to disk.                                   |
```

spacy_llm/tests/pipeline/test_llm.py
```diff
@@ -159,6 +164,9 @@ def parse_responses(
         ) -> Iterable[Doc]:
             return docs
 
+        def prompt_template(self) -> str:
+            return ""
+
```
svlandeg, line 169: I worry a little here that we're making it more complicated/extensive for users to implement a custom task. Now they'll always have to implement `prompt_template` as well, right?
Isn't there a way we can say that if this property is not available we just assume it to be `""`?

rmitsch: Done in https://github.com/explosion/spacy-llm/pull/192/commits/28d6d4b51a6c2c5da8f7a2f89181c5842511d751 and https://github.com/explosion/spacy-llm/pull/192/commits/cd1ccff222bcd498107dd53b5b97b68ed87abb3e.

spacy_llm/cache.py
```diff
@@ -62,31 +66,71 @@ def __init__(
...
+
+    def initialize(self, vocab: Vocab, task: LLMTask) -> None:
+        """
+        Initialize cache with data not available at construction time.
+        vocab (Vocab): Vocab object.
+        task (LLMTask): Task.
+        """
+        self._vocab = vocab
+        if isinstance(task, PromptTemplateProvider):
+            self.prompt_template = task.prompt_template
```
svlandeg, line 79: I think this is a nice solution :-)

spacy_llm/cache.py
```diff
@@ -62,31 +66,71 @@ def __init__(
...
+
+        # Store prompt template as file.
+        prompt_template_path = self._path / "prompt_template.txt"
+        # If no file exists: store in plain text for easier debugging.
+        if not prompt_template_path.exists():
+            with open(prompt_template_path, "w") as file:
+                file.write(self._prompt_template)
+
+        else:
+            # If file exists and prompt template is not equal to new prompt template: raise, as this indicate we are
```
svlandeg, line 113:
```suggestion
            # If file exists and prompt template is not equal to new prompt template: raise, as this indicates we are
```

svlandeg (approved):
Nice!

PR #195, opened 2023-06-28
Quote entities in NER task `spacy.NER.v3`

rmitsch:
## Description

Introduce task `spacy.NER.v3` with quoted strings to allow commas in entity names. Adjust parsing in `SpanTask` to accomodate for both quoted and unquoted entity names.

### Types of change

Enhancement.

## Checklist

  - all tests requiring a GPU

### DIFF
--- a/spacy_llm/tasks/util/span.py
+++ b/spacy_llm/tasks/util/span.py
@@ -1,3 +1,4 @@
+import shlex
 from typing import Callable, Dict, Iterable, List, Optional, Tuple, Type
 
 import jinja2
@@ -35,6 +36,7 @@ class SpanTask(SerializableTask[SpanExample]):
         self._alignment_mode = alignment_mode
         self._case_sensitive_matching = case_sensitive_matching
         self._single_match = single_match
+        self._uses_quoted_strings = "quoted strings" in template
 
     @property
     def labels(self) -> Tuple[str, ...]:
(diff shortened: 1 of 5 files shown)


spacy_llm/tasks/util/span.py
```diff
@@ -35,6 +36,7 @@ def __init__(
         self._alignment_mode = alignment_mode
         self._case_sensitive_matching = case_sensitive_matching
         self._single_match = single_match
+        self._uses_quoted_strings = "quoted strings" in template
```
rmitsch, line 39: This is hacky. The motivation is that parsing comma-delimited lists of either quoted or unquoted strings was cumbersome - parsing is easier if we know whether to expect quoted or unquoted strings, hence this check. Putting this up because I'm looking for better ideas (either to check whether quoted or unquoted strings are used or a unified approach to parsing the response).

kabirkhan:
Quotes feel a bit tough to manage and not all that uncommon in text. In this [wip PR](https://github.com/explosion/spacy-llm/pull/180/files) for Chain of Thought prompting, the `|` separator seems to work pretty well. Also in the Textcat task I added `==None==` as the label if none of the labels fits for some text. 

These kinds of separators seem pretty easy for the language model to reason about and are pretty easy to work with in the eventual parser. 

The `separator` can even be an optional argument with a default value here.

kabirkhan:
Also while we're making a `spacy.NER.v3` can we add a provision in the prompt like:

```
From the Text input provided, extract named entities for each label in the following format:
{# whitespace #}
{# whitespace #}
{%- for label in labels -%}
{{ label }}: <comma-delimited list of quoted strings>
{# whitespace #}
{%- endfor -%}
{# whitespace #}

Stick to this output format exactly with no newline characters between the label: <comma delimited list of strings>
```

to help with the scenario where the model outputs something like:


```
PERSON:
Kabir
```

instead of 


```
PERSON: Kabir
```

rmitsch:
I'm fine with a customizable separation char as well. It might also not be sensible to slightly modify the prompt now if we'll exchange it soon anyway. @vin-ivar Do you have an ETA in mind for the [CoT span task PR](https://github.com/explosion/spacy-llm/pull/180/files)?

vin-ivar:
The CoT PR already looks quite complete to be honest, and it shouldn't longer than until next Friday to wrap it up, so if you could wait that'd be ideal!

rmitsch:
Closing this in favor of https://github.com/explosion/spacy-llm/pull/180/files.

PR #197, opened 2023-07-04
Explicitly display error codes

vin-ivar:
## Description
Explicitly display the error code/reason the server returns, it helps to be a bit verbose when you're debugging, rather than throw generic errors.

### Types of change
Enhancement

## Checklist

  - all tests requiring a GPU


svlandeg (approved):
Good idea!

PR #201, opened 2023-07-05
[wip] add support for classification with Cohere models

vin-ivar:
## Description
Cohere has models that explicitly perform [text classification](https://docs.cohere.com/reference/classify); this workflow does not involve prompting, and as such is somewhat different to everything else in `spacy-llm`. It's unclear that this should be merged at all, but I'm making a PR in case anyone decides to pick this up later.

### Types of change
New feature

## Checklist

  - all tests requiring a GPU


rmitsch:
Closing due to being out of scope after internal discussion.

PR #209, opened 2023-07-10
Fix OpenLLaMA model names

rmitsch:
## Description

Fix OpenLLaMA model names after updates on the corresponding HuggingFace account.

### Types of change

Fix.

## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,6 +7,9 @@ ran. If your test fixes a bug reported in an issue, don't forget to include the
 issue number. If your PR is still a work in progress, that's totally fine – just
 include a note to let us know. -->
 
+### Corresponding documentation PR
+<!--- Add the link to the corresponding documentation PR here, if applicable. -->
+
 ### Types of change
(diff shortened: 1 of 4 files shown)


svlandeg:
I think we should release this as a bugfix release instead, because the OpenLlama models will remain broken otherwise until 0.5.0, right?

rmitsch:
> I think we should release this as a bugfix release instead of targeting develop, because the OpenLlama models will remain broken otherwise until 0.5.0, right?

Good point.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,6 +7,9 @@ ran. If your test fixes a bug reported in an issue, don't forget to include the
 issue number. If your PR is still a work in progress, that's totally fine – just
 include a note to let us know. -->
 
+### Corresponding documentation PR
+<!--- Add the link to the corresponding documentation PR here, if applicable. -->
+
```
svlandeg, line 12: In this case, you've made the edit directly on the spaCy PR?

svlandeg (approved):

PR #215, opened 2023-07-12
All defaults in config by manual registry resolution

rmitsch:
## Description

Always have defaults in config by not accepting `None` for mandatory arguments and instead accepting registry handles.

todo 

### Types of change

Enhancement.

## Checklist


spacy_llm/tasks/rel.py
```diff
@@ -65,7 +65,7 @@ def make_rel_task(
     template: str = _DEFAULT_REL_TEMPLATE,
     label_definitions: Optional[Dict[str, str]] = None,
     examples: ExamplesConfigType = None,
-    normalizer: Optional[Callable[[str], str]] = None,
+    normalizer: Union[Callable[[str], str], str] = "spacy.LowercaseNormalizer.v1",
```
rmitsch, line 68: Do we actually want to get rid of all the default values? I think usually we have default values in the factory functions, but not in the (sub-)components themselves.

svlandeg:
After internal discussion, we decided to abandon this route for now.

PR #222, opened 2023-07-17
Fix Falcon HF warnings

rmitsch:
## Description

Fix Falcon HF warnings.

## Checklist

### DIFF
--- a/spacy_llm/models/hf/falcon.py
+++ b/spacy_llm/models/hf/falcon.py
@@ -22,7 +22,12 @@ class Falcon(HuggingFace):
         self._device: Optional[str] = None
         super().__init__(name=name, config_init=config_init, config_run=config_run)
         assert isinstance(self._tokenizer, transformers.PreTrainedTokenizerBase)
-        self._config_run["eos_token_id"] = self._tokenizer.eos_token_id
+        # self._config_run["eos_token_id"] = self._tokenizer.eos_token_id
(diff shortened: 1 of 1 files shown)


spacy_llm/models/hf/falcon.py
```diff
@@ -22,7 +22,12 @@ def __init__(
         self._device: Optional[str] = None
         super().__init__(name=name, config_init=config_init, config_run=config_run)
         assert isinstance(self._tokenizer, transformers.PreTrainedTokenizerBase)
-        self._config_run["eos_token_id"] = self._tokenizer.eos_token_id
+        # self._config_run["eos_token_id"] = self._tokenizer.eos_token_id
```
svlandeg, line 25: Let's remove the code line entirely, instead of just commenting it out. If we do want to leave it in, we should add another comment explaining why it's there & commented out.

svlandeg (commented):
I'm still seeing other warnings when using Falcon, but this PR resolves the one about "modifying the pretrained model configuration to control generation" and the approach looks good to me! Yes let's do this for the other HF models, too.

svlandeg (approved):
I meant to approve :-)

spacy_llm/models/hf/falcon.py
rmitsch: Forgot to remove it. Removed in https://github.com/explosion/spacy-llm/pull/222/commits/ae86e6d3a6d40146b4a2f390a1e377b4845eaf16.

PR #233, opened 2023-07-26
Catch blocking errors from Cohere

ljvmiranda921:
Sometimes, Cohere returns a blocking error (e.g., blocked output or blocked input) because of some violation to their [Usage Agreement](https://docs.cohere.com/docs/usage-guidelines).  We want to catch those and give users the option to "ignore" it via the `self._strict` argument. The error is often returned as an `HTTPError`, so there's no way to catch it from the response.

### Types of change

Bug fix

## Checklist


rmitsch (approved):
It's whack that we have to look for markers in their error messages, but not much we can do about that.

Thanks for this!

PR #236, opened 2023-07-31
increase default max_tokens for older non-chat OpenAI models so NER/spancat works

kabirkhan:
## Description

The default `max_tokens` for the LLM response of the old completions endpoint for OpenAI is `16` tokens. This often causes too short of an output for the NER/SpanCat tasks (and of course for longer tasks like summarization). 

Setting the default at the API level higher value is a probably a good bet as a default but we should also probably update the docs here. 

This is only required for the legacy models, the chat completion models set this to Infinity by default. 

https://github.com/explosion/spaCy/pull/12961

### Types of change

enhancement

## Checklist


spacy_llm/models/rest/openai/registry.py
```diff
@@ -6,10 +6,12 @@
 from ....registry import registry
 from .model import Endpoints, OpenAI
 
+DEFAULT_TEMPERATURE = 0.0
```
rmitsch, line 9: Let's prefix this with `_`, as it's not used anywhere else.

rmitsch (commented):
This seems reasonable overall. Some notes:
- You mention setting `max_tokens` to 100, but the actual values are >2k. Why the divergence between description and code?
- Have you run the external tests with this change?
- The docs on spacy.io should be updated as well.

spacy_llm/models/rest/openai/registry.py
```diff
@@ -45,7 +47,7 @@ def openai_gpt_4(
 
 @registry.llm_models("spacy.GPT-3-5.v1")
 def openai_gpt_3_5(
-    config: Dict[Any, Any] = SimpleFrozenDict(),
+    config: Dict[Any, Any] = SimpleFrozenDict(temperature=_DEFAULT_TEMPERATURE),
```
rmitsch, line 50: No `max_tokens` for GPT-3.5 is intentional?

spacy_llm/models/rest/openai/registry.py
```diff
@@ -6,10 +6,12 @@
 from ....registry import registry
 from .model import Endpoints, OpenAI
 
+_DEFAULT_TEMPERATURE = 0.0
+
 
 @registry.llm_models("spacy.GPT-4.v1")
 def openai_gpt_4(
-    config: Dict[Any, Any] = SimpleFrozenDict(),
+    config: Dict[Any, Any] = SimpleFrozenDict(temperature=_DEFAULT_TEMPERATURE),
```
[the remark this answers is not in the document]
kabirkhan: yep for the Chat completion endpoint `max_tokens` defaults to `infinity`

kabirkhan:
External tests all run correctly here but yeah docs need to be updated still, will add a link to a PR shortly

svlandeg (commented):
When we change the default settings like this for these models, don't we have to bump all the versions?

rmitsch:
> When we change the default settings like this for these models, don't we have to bump all the versions?

Hm...I'd say we don't have to, but we mention it in the release notes.

svlandeg:
> > When we change the default settings like this for these models, don't we have to bump all the versions?
> 
> Hm...I'd say we don't have to, but we mention it in the release notes.

Hmm. Then at the very least we should have this on `develop` and not `main`?

rmitsch:
> Hmm. Then at the very least we should have this on develop and not main?

Fair, changed to `develop`.

rmitsch:
Added a [docs PR](https://github.com/explosion/spaCy/pull/12903). I think we can merge both the docs PR and this one?

svlandeg:
It looks like the external tests are failing with an error about writing to a frozen dict...

svlandeg:
The only remaining filaing external test should be fixed with the new `confection` release.

spacy_llm/models/rest/openai/registry.py
```diff
@@ -43,6 +67,45 @@ def openai_gpt_4(
...
+        or other response object that does not conform to the expectation of how a well-formed response object from
+        this API should look like). If False, the API error responses are returned by __call__(), but no error will
+        be raised.
+    max_tries (int): Max. number of tries for API request.
+    interval (float): Time interval (in seconds) for API retries in seconds. We implement a base 2 exponential backoff
+        at each retry.
+    max_request_time (float): Max. time (in seconds) to wait for request to terminate before raising an exception.
+    RETURNS (Callable[[Iterable[str]], Iterable[str]]]): OpenAI instance for 'gpt-3.5' model using REST to
+        prompt API.
+    """
```
rmitsch, line 97: Do we want to remove the docstrings and maybe just describe the params once at the top of the file? I feel like this is just a lot of bloat.

svlandeg: I agree, I wanted to do something similar as well, just wasn't sure about the format.

svlandeg:
new docs PR: https://github.com/explosion/spaCy/pull/12961

spacy_llm/models/rest/openai/registry.py
svlandeg: An argument against that would be if users would like to call these functions directly in Python code.

rmitsch: Yeah, it's not great either way :thinking:

svlandeg: https://github.com/explosion/spacy-llm/pull/236/commits/21bef95b16ba8747eea0b6563e8161362d578f68

rmitsch: Good with me.

rmitsch (approved):

PR #242, opened 2023-08-03
Swappable parsing

rmitsch:
## Description

Make parsing a swappable, configurable attribute to allow modify parsing behaviour without retiring the entire task class.

todo

### Types of change

New feature.

## Checklist


spacy_llm/tasks/lemma.py
```diff
@@ -90,6 +95,11 @@ def generate_prompts(self, docs: Iterable[Doc]) -> Iterable[str]:
 
     def parse_responses(
         self, docs: Iterable[Doc], responses: Iterable[str]
+    ) -> Iterable[Doc]:
+        return self._parse_responses(docs, responses)
+
+    def _parse_responses_default(
```
rmitsch, line 101: This would be temporary. If we decide to do this, all response parse functions would be registered.

spacy_llm/ty.py
```diff
@@ -16,6 +16,7 @@
 _Response = Any
 
 PromptExecutor = Callable[[Iterable[_Prompt]], Iterable[_Response]]
+TaskResponseParser = Callable[[Iterable[Doc], Iterable[_Response]], Iterable[Doc]]
```
kabirkhan, line 19: this is complex enough for a protocol and I think `PromptExecutor` should probably be one as well.

rmitsch: Hm, I don't see the benefit of that. IIRC protocols are tricky w.r.t. typing if only a function (and not a callable object) is supplied. Independent from that I see protocols only making sense for classes expected to have at least one method with a certain signature. IMO a type definition is the simplest tool for the job here.

I'm interested in hearing your arguments in favor of using protocols however, I might be missing something.

rmitsch:
It might make sense to move each task into their own module and split the actual task implementation from the factory methods and utils (such as the parsing methods). Over time this will become unwieldy otherwise.

rmitsch:
Making parsing swappable is a bit more complicated for tasks inheriting from `SpanTask`, as there other class methods are invoked. There isn't a strong reason for that though, this setup works fine with moving all parsing-related functions out of the class and registering them (which is what this PR aims for, ultimately).

svlandeg (commented):
This looks like a promising approach to me to allow more flexibility both for users, as well as for our own builtin functionality.

As discussed internally, we'd also need to make the `_create_prompt_example` functionality swappable.

svlandeg:
> It might make sense to move each task into their own module and split the actual task implementation from the factory methods and utils (such as the parsing methods). Over time this will become unwieldy otherwise.

We can definitely do that, yep.

spacy_llm/ty.py
svlandeg: I agree with Raphael that a type definition seems fine here.

spacy_llm/tasks/lemma/registry.py
```diff
@@ -0,0 +1,47 @@
+from typing import Optional, Type
+
+from ...registry import registry
+from ...ty import ExamplesConfigType, FewshotExample, TaskResponseParser
+from .task import LemmaTask
+from .util import DEFAULT_LEMMA_TEMPLATE_V1, LemmaExample, parse_responses_v1
```
svlandeg, line 6: I think `util` is not a good name for this, as it's not just boilerplate utility code, this is specific functionality. Maybe a `parser.py` and `examples.py`. And read the templates in `task.py` (not married to any of these names/decisions though)

spacy_llm/tasks/lemma/registry.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+@registry.llm_misc("spacy.LemmaParser.v1")
+def make_lemma_parser() -> TaskResponseParser:
+    return parse_responses_v1
+
+
+@registry.llm_misc("spacy.LemmaFewshotExample.v1")
+def make_lemma_fewshot_example() -> TaskResponseParser:
+    return parse_responses_v1
```
svlandeg, line 16: Are these supposed to be the same?

spacy_llm/tasks/lemma/task.py
```diff
@@ -0,0 +1,90 @@
...
+
+from ...ty import FewshotExample, TaskResponseParser
+from ..util import ExampleType, SerializableTask
+from .util import DEFAULT_LEMMA_TEMPLATE_V1
+
+
+class LemmaTask(SerializableTask[ExampleType], Generic[ExampleType]):
+    def __init__(
+        self,
+        template: str = DEFAULT_LEMMA_TEMPLATE_V1,
```
svlandeg, line 17: I think we should remove the default template here and assume it's always defined?

spacy_llm/tasks/lemma/task.py
```diff
@@ -0,0 +1,90 @@
...
+        parse_responses: Optional[TaskResponseParser] = None,
```
svlandeg, line 18: I don't think this can be `Optional`? It'd have to be defined, right?

spacy_llm/tasks/lemma/task.py
```diff
@@ -0,0 +1,90 @@
...
+        examples: Optional[List[ExampleType]] = None,
+    ):
+        """Default lemmatization task.
+
+        template (str): Prompt template passed to the model.
+        parse_responses (Optional[TaskResponseParser]): Callable for parsing LLM responses for this task.
+        examples (Optional[Callable[[], Iterable[Any]]]): Optional callable that
+            reads a file containing task examples for few-shot learning. If None is
+            passed, then zero-shot learning will be used.
+        """
```
svlandeg, line 29: `fewshot_example_type` isn't documented in the docstring

spacy_llm/tasks/lemma/task.py
```diff
@@ -0,0 +1,90 @@
...
+from ..util import ExampleType, SerializableTask
+from .util import DEFAULT_LEMMA_TEMPLATE_V1
+
+
+class LemmaTask(SerializableTask[ExampleType], Generic[ExampleType]):
+    def __init__(
+        self,
+        template: str = DEFAULT_LEMMA_TEMPLATE_V1,
+        parse_responses: Optional[TaskResponseParser] = None,
+        fewshot_example_type: Optional[Type[FewshotExample]] = None,
```
svlandeg, line 19: Again, not sure how this can be `Optional` if the code asserts it to be not `None`?

svlandeg (commented):
I like this direction! Leaving some first comments.

spacy_llm/tasks/lemma/task.py
rmitsch: We do have duplicated default values for all tasks. Duplicated as in default values are defined in the factory function _and_ the task. Should we just remove all default values from all task constructors?

spacy_llm/tasks/lemma/task.py
rmitsch: It's set in `__init__()`.

spacy_llm/tasks/lemma/task.py
rmitsch: Fixed in https://github.com/explosion/spacy-llm/pull/242/commits/047a496ca863c972e2a35600b6492960cac81706.

spacy_llm/tasks/lemma/registry.py
rmitsch: The latter is redundant, forgot to remove it. Fixed in https://github.com/explosion/spacy-llm/pull/242/commits/047a496ca863c972e2a35600b6492960cac81706.

spacy_llm/tasks/lemma/registry.py
rmitsch: Done in https://github.com/explosion/spacy-llm/pull/242/commits/bd416bafc2bf4461791dc666c790502af9dc430e.

spacy_llm/tasks/lemma/task.py
svlandeg: Urgh. Right.

spacy_llm/tasks/lemma/task.py
svlandeg: Hm, ok. Should we move the `assert` to `__init__()` though?

spacy_llm/tasks/lemma/task.py
svlandeg: I think so? Like in spaCy's factories we also only define the defaults in the `default_config`'s, right?

spacy_llm/tasks/rel.py
```diff
@@ -1,7 +1,13 @@
...
 
 import jinja2
-from pydantic import BaseModel, Field, ValidationError, validator
+
+try:
+    from pydantic.v1 import validator
+except ImportError:
+    from pydantic import validator
+
+from pydantic import BaseModel, Field, ValidationError
```
adrianeboyd, line 10: I think you want to move all these imports into the `try`/`except` block because they might not interact like you'd expect? (I am not 100% sure, but it could be the source of the warnings. I don't understand why it's only in some python versions, though.)

rmitsch: I'm also wondering why it's only on Mac with Python 3.8 :upside_down_face: Done in https://github.com/explosion/spacy-llm/pull/242/commits/3bcf2a8339439cc268cee055887c3fec3aa5b6c4.

spacy_llm/tasks/lemma/task.py
rmitsch: Done in https://github.com/explosion/spacy-llm/pull/242/commits/0f869a729b46ffbfb9e23fcf57b63be1ce1adbd4.

spacy_llm/tasks/lemma/task.py
rmitsch: From what I've seen (e. g. [`EntityLinker`](https://github.com/explosion/spaCy/blob/master/spacy/pipeline/entity_linker.py#L168)), yes.

spacy_llm/tasks/rel.py
rmitsch: Done in https://github.com/explosion/spacy-llm/pull/242/commits/19410177bd8630c55ed1f58529cf4c9091fb3a02. Seems to get rid of the warnings indeed - good call, thank you!

spacy_llm/tasks/lemma/registry.py
```diff
@@ -0,0 +1,43 @@
...
+    if fewshot_example_type is None:
+        fewshot_example_type = LemmaExample
+    lemma_examples = (
+        [fewshot_example_type(**eg) for eg in raw_examples] if raw_examples else None
+    )
+
+    return LemmaTask(
+        template=template,
+        parse_responses=parse_responses if parse_responses else parse_responses_v1,
+        fewshot_example_type=fewshot_example_type,
```
svlandeg, line 41: I suggest to either shorten both (as below), or spell out both before the `return` statement:

```suggestion
        parse_responses=parse_responses or parse_responses_v1,
        fewshot_example_type=fewshot_example_type or LemmaExample,
```

I wouldn't use the `or` style for more complicated things like the `lemma_examples` definition though.

spacy_llm/tasks/lemma/task.py
```diff
@@ -0,0 +1,96 @@
...
+from spacy.language import Language
+from spacy.scorer import Scorer
+from spacy.tokens import Doc
+from spacy.training import Example
+
+from ...ty import FewshotExample, TaskResponseParser
+from ..templates import read_template
+from ..util import ExampleType, SerializableTask
+
+DEFAULT_LEMMA_TEMPLATE_V1 = read_template("lemma")
```
svlandeg, line 13: It bothers me that we didn't version the v1's in their name, can we still change that for the sake of consistency?

spacy_llm/tasks/lemma/task.py
```diff
@@ -0,0 +1,96 @@
...
+            reads a file containing task examples for few-shot learning. If None is
+            passed, then zero-shot learning will be used.
+        """
+        self._template = template
+        self._parse_responses = parse_responses
+        self._prompt_examples = examples or []
+        self._fewshot_example_type = fewshot_example_type
+
+        assert callable(self._parse_responses)
+        assert self._fewshot_example_type is not None
```
svlandeg, line 39: Let's discuss this a bit more internally - it feels weird to type these as `Optional` and then immediately `assert` that they're not `None`. It's all related to the serialization & needing `None` defaults.

spacy_llm/tasks/lemma/task.py
```diff
@@ -0,0 +1,96 @@
...
+        """Initializes prompt examples from Doc examples.
+        get_examples (Callable[[], Iterable["Example"]]): Callable that provides examples
+            for initialization.
+        nlp (Language): Language instance.
+        n_prompt_examples (int): How many prompt examples to infer from the provided Example objects.
+            0 by default. Takes all examples if set to -1.
+        """
+        for eg in get_examples():
+            if n_prompt_examples < 0 or len(self._prompt_examples) < n_prompt_examples:
+                assert self._fewshot_example_type
```
svlandeg, line 57:
```suggestion
```

spacy_llm/tasks/lemma/task.py
```diff
@@ -0,0 +1,96 @@
...
+            prompt = _template.render(
+                text=doc.text,
+                examples=self._prompt_examples,
+            )
+            yield prompt
+
+    def parse_responses(
+        self, docs: Iterable[Doc], responses: Iterable[str]
+    ) -> Iterable[Doc]:
+        assert callable(self._parse_responses)
```
svlandeg, line 77:
```suggestion
```
(moved to `__init__`)

spacy_llm/tasks/lemma/task.py
```diff
@@ -0,0 +1,96 @@
...
+        return self._parse_responses(docs, responses)
+
+    def scorer(
+        self,
+        examples: Iterable[Example],
+    ) -> Dict[str, Any]:
+        """Scores lemmatization accuracy on provided examples.
+        examples (Iterable[Example]): Examples to determine score against.
+        """
+        return Scorer.score_token_attr(examples, "pos")
```
svlandeg, line 87: Wait why isn't this `"lemma"`?

spacy_llm/tasks/lemma/task.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @property
+    def _cfg_keys(self) -> List[str]:
+        return ["_template"]
+
+    @property
+    def _Example(self) -> Type[FewshotExample]:
+        assert self._fewshot_example_type is not None
```
svlandeg, line 95:
```suggestion
```

spacy_llm/tasks/lemma/task.py
rmitsch: Needed for `mypy` to be happy, unfortunately. Alternatively I can add a `mypy` ignore comment.

rmitsch: Actually, this should be rendered redundant anyway once this isn't optional anymore.

spacy_llm/tasks/lemma/task.py
rmitsch: Good catch :smile:

spacy_llm/tasks/lemma/task.py
rmitsch: Fixed in https://github.com/explosion/spacy-llm/pull/242/commits/46fc5cfc0f498ef54a68a3910e5c0653cbcf20dd.

spacy_llm/tasks/lemma/task.py
rmitsch: Fixed in https://github.com/explosion/spacy-llm/pull/242/commits/46fc5cfc0f498ef54a68a3910e5c0653cbcf20dd.

spacy_llm/tasks/lemma/task.py
rmitsch: Fixed in https://github.com/explosion/spacy-llm/pull/242/commits/46fc5cfc0f498ef54a68a3910e5c0653cbcf20dd.

spacy_llm/tasks/lemma/task.py
rmitsch: Fixed in https://github.com/explosion/spacy-llm/pull/242/commits/46fc5cfc0f498ef54a68a3910e5c0653cbcf20dd.

spacy_llm/tasks/lemma/task.py
rmitsch: Updated in https://github.com/explosion/spacy-llm/pull/242/commits/43368c3f94f11d97f2e2efa94205a201dc8d6129.

rmitsch:
Failing tests are expected at this point, as I've started, but haven't finished, working on tasks derived from `SpanTask`.

spacy_llm/tasks/ner/registry.py
```diff
@@ -0,0 +1,115 @@
...
+    case_sensitive_matching: bool = False,
+    single_match: bool = False,
+):
+    """NER.v1 task factory.
+
+    parse_responses (Optional[TaskResponseParser]): Callable for parsing LLM responses for this task.
+    fewshot_example_type (Optional[Type[FewshotExample]]): Type to use for fewshot examples.
+    labels (str): Comma-separated list of labels to pass to the template.
+        Leave empty to populate it at initialization time (only if examples are provided).
+    template (str): Prompt template passed to the model.
```
svlandeg, line 29:
```suggestion
```

spacy_llm/tasks/ner/registry.py
```diff
@@ -0,0 +1,115 @@
...
+    label_definitions (Optional[Dict[str, str]]): Map of label -> description
+        of the label to help the language model output the entities wanted.
+        It is usually easier to provide these definitions rather than
+        full examples, although both can be provided.
```
svlandeg, line 33:
```suggestion
```

svlandeg (commented):
Mostly this looks good!

Let's 
- Fix the merge conflict
- Double check the docs strings for consistency
- get the refactor PR merged in

And then I can do a final review round?

spacy_llm/tasks/span/parser.py
```diff
@@ -0,0 +1,67 @@
...
+            if norm_label in label_dict:
+                # Get the phrases / spans for each entity
+                if phrases.strip():
+                    _phrases = [p.strip() for p in phrases.strip().split(",")]
+                    output.append((label_dict[norm_label], _phrases))
+
+    return output
+
+
+def parse_responses(responses: Iterable[str], **kwargs) -> Iterable[Span]:
```
kabirkhan, line 35: I don't think `kwargs` makes sense here. It actually loses quite a bit of clarity in the function. I'm assuming this is done so we have a single `TaskResponseParserProtocol`.

If you're looking at this function in isolation, we don't validate `kwargs` at all, we just assume these keys are present. This "works" but is not very easy to understand and is a source of potential bugs.

spacy_llm/tasks/ner/task.py
```diff
@@ -0,0 +1,118 @@
...
+DEFAULT_NER_TEMPLATE_V2 = read_template("ner.v2")
+
+
+class NERTask(SpanTask):
+    def __init__(
+        self,
+        labels: List[str],
+        template: str,
+        parse_responses: TaskResponseParserProtocol,
+        fewshot_example_type: Type[SpanExample],
```
kabirkhan, line 24: This type can't be `SpanExample` because `SpanExample` assumes the `text`, and `entities` fields. The COT prompting has different types for the keys and wouldn't be compatible.

spacy_llm/tasks/span/parser.py
rmitsch: I agree on the merits of explicitly stating arguments over `kwargs`. I originally went for the latter because I couldn't find another way to make the types of `_parse_responses` agree with the parent's `__init__()` signature. I'll have another look into how to at least approximate a typing match there.

spacy_llm/tasks/ner/task.py
rmitsch: In this case `SpanTask` will have to be modified as well, as it relies on `entities` and `text` being attributes of `SpanExample`. I suggest approaching this in the CoT PR though.

spacy_llm/tasks/ner/task.py
```diff
@@ -0,0 +1,118 @@
...
+        labels (List[str]): List of labels to pass to the template.
+            Leave empty to populate it at initialization time (only if examples are provided).
+        template (str): Prompt template passed to the model.
+        parse_responses (TaskResponseParser): Callable for parsing LLM responses for this task.
+        fewshot_example_type (Type[FewshotExample]): Type to use for fewshot examples.
+        label_definitions (Optional[Dict[str, str]]): Map of label -> description
+            of the label to help the language model output the entities wanted.
+            It is usually easier to provide these definitions rather than
+            full examples, although both can be provided.
+        examples (Optional[Callable[[], Iterable[Any]]]): Optional callable that
```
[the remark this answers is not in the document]
rmitsch: Do we want to call this attribute `prompt_examples` or `fewshot_examples` or `examples`? This is a inconsistent currently.

spacy_llm/tasks/span/parser.py
kabirkhan: Check out `ParamSpec`, that's the general way to type a callable and say inherit these arguments

spacy_llm/ty.py
```diff
@@ -55,46 +66,79 @@ def from_disk(
...
     ) -> Dict[str, Any]:
         """Scores performance on examples."""
-        ...
 
 
 @runtime_checkable
-class LLMTask(Protocol):
-    def generate_prompts(self, docs: Iterable[Doc]) -> Iterable[_Prompt]:
-        ...
+class CallableScorableProtocol(Protocol):
```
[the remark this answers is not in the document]
rmitsch: We have `ScorerProtocol` already. These two are used for different purposes. I'm open to renaming suggestions.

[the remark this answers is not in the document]
rmitsch: Eh, I meant we have `ScorableProtocol` already. The difference between `ScorableProtocol` and `ScorerProtocol` isn't significant enough IMO.

spacy_llm/tasks/span/parser.py
rmitsch: Good point, thanks. Not sure if after transitioning to explicit signatures for the concrete `parse_responses()` functions using `ParamSpec` over `Callable[..., Iterable[Any]]` makes a difference though, will look into it.

rmitsch:
Ready for a final review round. We should decide whether we want to keep the file structure for tasks as-is - we could also merge `parser.py`, `example.py`, `scorer.py` into a single file?

spacy_llm/tasks/ner/task.py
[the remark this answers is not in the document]
svlandeg: Definitely not `examples`. I thought the current name was `prompt_examples`? Though I'm happy with `fewshot_examples`, too.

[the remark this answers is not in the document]
rmitsch: Some were `prompt_examples`, some were `examples`. I'll go with `fewshot_examples` then, I think it's the most recognizable term.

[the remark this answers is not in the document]
rmitsch: Done in https://github.com/explosion/spacy-llm/pull/242/commits/b443670e1a7c407fcb637e32b9088b8b9b5045b7. However, changing `examples` to `fewshot_examples` also in the factory functions (as it's currently the case in this PR) is a breaking change. Bumping task versions just for this also seems like a bother. 

Should we revert `fewshot_examples` to `examples` in the task factory functions?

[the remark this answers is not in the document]
svlandeg: As discussed internally, let's keep the registered task definitions as-is to avoid too many breaking changes in existing configs, and to minimize the refactoring done in this PR, which is difficult to track diff-wise.

rmitsch:
`scorer` currently runs on `**kwargs` as well - we should update that to a generic protocol in a follow-up PR. Added an internal ticket.

spacy_llm/ty.py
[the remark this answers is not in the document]
kabirkhan: What about `ScorerProtocol` for the function implementation and `ScorableTaskProtocol` for if the task implements the scorer method?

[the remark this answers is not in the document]
rmitsch: Sounds good, changed in https://github.com/explosion/spacy-llm/pull/242/commits/aee012260cabe19625e7eb7f851e8cf9b2530ff4.

spacy_llm/tasks/spancat/registry.py
```diff
@@ -0,0 +1,123 @@
...
+    scorer: Optional[Scorer] = None,
+):
+    """SpanCat.v1 task factory.
+
+    parse_responses (Optional[TaskResponseParserProtocol[SpanCatTask]]): Callable for parsing LLM responses for this
+        task.
+    fewshot_example_type (Optional[Type[FewshotExample]]): Type to use for fewshot examples.
+    labels (str): Comma-separated list of labels to pass to the template.
+        Leave empty to populate it at initialization time (only if examples are provided).
+    template (str): Prompt template passed to the model.
```
svlandeg, line 32:
```suggestion
```

svlandeg: (assuming that this wasn't an option to define in v1)

spacy_llm/tasks/spancat/task.py
```diff
@@ -0,0 +1,112 @@
...
+        normalizer: Optional[Callable[[str], str]],
+        alignment_mode: Literal["strict", "contract", "expand"],
+        case_sensitive_matching: bool,
+        single_match: bool,
+        scorer: Scorer,
+    ):
+        """Default SpanCat task.
+
+        parse_responses (TaskResponseParser): Callable for parsing LLM responses for this task.
+        fewshot_example_type (Type[FewshotExample]): Type to use for fewshot examples.
```
svlandeg, line 35: I think there's one inconsistency left here after all the renaming back-and-forth: `prompt_example_type` vs. `fewshot_example_type` (throughout the PR)

svlandeg: Let's stick with `prompt_example_type` for now?

spacy_llm/tasks/summarization/parser.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def parse_responses_v1(
+    task: SummarizationTask, docs: Iterable[Doc], responses: Iterable[str]
+) -> Iterable[str]:
+    """Parses LLM responses for spacy.Summarization.v1.
+    task (SummarizationTask): Task instance.
+    docs (Iterable[Doc]): Corresponding Doc instances.
+    responses (Iterable[str]): LLM responses.
+    field (str): Field to store responses in.
```
svlandeg, line 15:
```suggestion
```

spacy_llm/tasks/textcat/parser.py
```diff
@@ -0,0 +1,46 @@
...
+from spacy.tokens import Doc
+from wasabi import msg
+
+from .task import TextCatTask
+
+
+def parse_responses_v1_v2_v3(
+    task: TextCatTask, docs: Iterable[Doc], responses: Iterable[str]
+) -> Iterable[Dict[str, float]]:
+    """Parses LLM responses for spacy.TextCat.v1.
```
svlandeg, line 12:
```suggestion
    """Parses LLM responses for spacy.TextCat.v1, v2 and v3
```

spacy_llm/ty.py
```diff
@@ -55,24 +58,69 @@ def from_disk(
...
+    def generate(cls, example: Example, **kwargs) -> Self:
+        """Create a fewshot example from a spaCy example.
+        example (Example): spaCy example.
+        """
+
+
 @runtime_checkable
-class Scorable(Protocol):
+class ScorableTask(Protocol):
+    """Differs from ScorableCallableProtocol in that it describes an object with a scorer() function, i. e. a scorable
```
svlandeg, line 72: What is `ScorableCallableProtocol`?

spacy_llm/ty.py
```diff
@@ -55,24 +58,69 @@ def from_disk(
...
         self,
         examples: Iterable[Example],
     ) -> Dict[str, Any]:
         """Scores performance on examples."""
-        ...
+
+
+@runtime_checkable
+class Scorer(Protocol):
+    """Differs from ScorableProtocol in that it describes a Callable with a call signature matching the scorer()
```
svlandeg, line 85: ScorableTask?

spacy_llm/tasks/builtin_task.py
```diff
@@ -0,0 +1,324 @@
...
+class BuiltinTask(abc.ABC):
+    """Abstract base task implementing interfaces and/or functionality expected from all built-in tasks:
+        - working prompt template strings
+        - swappable response parsers
+        - swappable prompt example type
+        - integration of fewshot example into the fully rendered prompt
+        - initializable (in line with other spaCy components)
+        - (de-)serialization
+
+    On the relation of BuiltinTask to LLMTaskProtocol: the latter specifies the minimal contract a task implementation
```
svlandeg, line 24: ctrl-F and replace:
- LLMTaskProtocol
- TaskResponseParserProtocol

svlandeg (commented):
This is looking good! Left a few final comments & fixes, nothing major.
Feel free to merge afterwards so we can move on to the CoT to put this in practice 🎉

svlandeg:
And maybe to be sure, run the GPU & external tests at the end before merging.

spacy_llm/tasks/builtin_task.py
rmitsch: Fixed in https://github.com/explosion/spacy-llm/pull/242/commits/b687d54a5b3f72dfd05f848e490c140b45d70a3b.

spacy_llm/ty.py
rmitsch: Fixed in https://github.com/explosion/spacy-llm/pull/242/commits/b687d54a5b3f72dfd05f848e490c140b45d70a3b.

spacy_llm/ty.py
rmitsch: Fixed in https://github.com/explosion/spacy-llm/pull/242/commits/b687d54a5b3f72dfd05f848e490c140b45d70a3b.

spacy_llm/tasks/lemma/registry.py
```diff
@@ -0,0 +1,49 @@
...
+    scorer: Optional[Scorer] = None,
+):
+    """Lemma.v1 task factory.
+
+    template (str): Prompt template passed to the model.
+    parse_responses (Optional[TaskResponseParser]): Callable for parsing LLM responses for this task.
+    fewshot_example_type (Optional[Type[FewshotExample]]): Type to use for fewshot examples.
+    examples (Optional[Callable[[], Iterable[Any]]]): Optional callable that reads a file containing task examples for
+        few-shot learning. If None is passed, then zero-shot learning will be used.
+    scorer (Optional[BuiltinScorableProtocol]): Scorer function.
```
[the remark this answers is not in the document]
rmitsch: Fixed in https://github.com/explosion/spacy-llm/pull/242/commits/b687d54a5b3f72dfd05f848e490c140b45d70a3b.

spacy_llm/tasks/spancat/task.py
rmitsch: Updated in https://github.com/explosion/spacy-llm/pull/242/commits/b804c5b48bf8b69f1271844ea77d88537a6bc07f.

spacy_llm/tasks/builtin_task.py
```diff
@@ -0,0 +1,324 @@
...
+        for key, value in cfg.items():
+            setattr(self, key, value)
+
+    def _get_prompt_examples(self) -> List[Dict[str, Any]]:
+        """Serialize examples."""
+        examples = [eg.dict() for eg in self._prompt_examples]
+        return examples
+
+    def _set_prompt_examples(self, examples: List[Dict[str, Any]]) -> None:
+        """Deserialize examples from bytes.
```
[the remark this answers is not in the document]
rmitsch: Fixed in https://github.com/explosion/spacy-llm/pull/242/commits/b804c5b48bf8b69f1271844ea77d88537a6bc07f.

svlandeg (approved):
🎉

PR #259, opened 2023-08-09
fix some of the generic typing

kabirkhan:
## Description

Cleaning up some of the generic typing.


rmitsch (commented):
What's the advantage of using `FewshotExampleType` outside of `ty.py`? That type was introduced only for the TypeVar with generic self/class approach to the factory function in `FewshotExample`. `Type[FewshotExample` is clearer to use everywhere else in my opinion.

rmitsch:
Alright, I removed `FewshotExampleType` in lieu of `Self` (wasn't aware this exists!). Closing this PR, since the remaining changes center around `FewshotExampleType`.

Thank you!

PR #264, opened 2023-08-15
Idea for generic response parser protocol

kabirkhan:
## Description

Pass the task instance as the first argument to the configurable parser. Type checking should pass with this.

## Checklist

- [ ] I confirm that I have the right to submit this contribution under the project's MIT license.


spacy_llm/tasks/span/parser.py
```diff
@@ -44,22 +49,22 @@ def parse_responses(responses: Iterable[str], **kwargs) -> Iterable[Span]:
...
         ):
             # For each phrase, find the substrings in the text
             # and create a Span
             offsets = find_substrings(
                 doc.text,
                 phrases,
-                case_sensitive=kwargs["case_sensitive_matching"],
-                single_match=kwargs["single_match"],
+                case_sensitive=task._case_sensitive_matching,
+                single_match=task._single_match,
```
kabirkhan, line 63: We're accessing private vars on the task now so that's the one downside (we'd probably need to make properties for our build-in tasks for these.

spacy_llm/ty.py
```diff
@@ -120,6 +112,16 @@ def parse_responses(
...
 
 
+_Task_contra = TypeVar("_Task_contra", bound=LLMTaskProtocol, contravariant=True)
+
+
+class TaskResponseParserProtocol(Protocol[_Task_contra]):
+    def __call__(
+        self, task: _Task_contra, docs: Iterable[Doc], responses: Iterable[Any]
+    ) -> Iterable[Any]:
+        ...
```
rmitsch, line 122: What's the benefit of this over
```suggestion

class TaskResponseParserProtocol:
    def __call__(
        self, task: LLMTaskProtocol, docs: Iterable[Doc], responses: Iterable[Any]
    ) -> Iterable[Any]:
        ...
```
?

rmitsch: (Also `docs: Iterable[Doc]` shouldn't be in there, but that's a detail)

rmitsch: Ah, because we have a generic protocol in this case. Got it.

PR #277, opened 2023-09-01
add_label functionality

svlandeg:
## Description
Implemented `add_label` functionality both for the `llm` component as well as for `BuiltinTaskWithLabels`, to be more in line with spaCy's other components that have labels.

This is mainly to support the use-case of using `spacy-llm` directly in Python code and not through the config.

TODO (soliciting feedback first)

### Types of change
Enhancement

## Checklist


svlandeg:
It seems like the external test suite is failing over wrong predictions - we should make this more robust.

spacy_llm/tasks/builtin_task.py
```diff
@@ -312,6 +312,14 @@ def _extract_labels_from_example(self, example: Example) -> List[str]:
     def labels(self) -> Tuple[str, ...]:
         return tuple(self._label_dict.values())
 
+    def add_label(self, label: str) -> int:
+        if not isinstance(label, str):
+            raise ValueError(Errors.E187)
+        if label in self.labels:
+            return 0
+        self._label_dict[self._normalizer(label)] = label
+        return 1
```
rmitsch, line 321: Why do we return 1 here?

svlandeg: To be in line with the `TrainablePipe.add_pipe` API: https://spacy.io/api/pipe#add_label

> RETURNS 0 if the label is already present, otherwise 1.

spacy_llm/tasks/builtin_task.py
```diff
@@ -312,6 +312,14 @@ def _extract_labels_from_example(self, example: Example) -> List[str]:
     def labels(self) -> Tuple[str, ...]:
         return tuple(self._label_dict.values())
 
+    def add_label(self, label: str) -> int:
+        if not isinstance(label, str):
+            raise ValueError(Errors.E187)
```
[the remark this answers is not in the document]
svlandeg: Pretty sure I copied this from spaCy, it might be that we used to allow `int`'s and then we'd put them through the vocab ourselves to do int->str conversion. Either way I figured it didn't hurt to have the same check as in `TrainablePipe` 🤷

PR #279, opened 2023-09-01
Simplify test suite

svlandeg:
## Description
Simplify test suite to make it less dependent on the actual answers from the LLM.
- Removed unnecessary label "DESTINATION" from the NER tests
- Updated the "multiple-entities-with-same-string" Paris test (too difficult for the LLM to get right)
- Updated the "correct prediction" test to be more simple & robust against odd LLM answers
- Updated the "comma" test to be more robust 

Also changed `"xx"` everywhere to `"en"` when we're testing on English sentences.

NA

### Types of change
test suite 

## Checklist


spacy_llm/tests/tasks/test_ner.py
```diff
@@ -221,52 +220,27 @@ def test_ner_config(config: Config):
...
         (
             "Marc and Bob both live in Ireland.",
             [("Marc", "PER"), ("Bob", "PER"), ("Ireland", "LOC")],
         ),
-        # same entity names
-        (
-            "Paris, who is a person, travelled to Paris, which is a city, to spend time with her friend Paris, who "
-            "is a person.",
-            [("Paris", "PER"), ("Paris", "LOC"), ("Paris", "PER")],
-        ),
```
svlandeg, line 239: I'm removing it here because getting the LLM to predict it correctly, is too uncertain. We already have the test `test_regression_span_task_response_parse` which checks that the parsing works, given the right answer.

rmitsch (approved):

PR #306, opened 2023-09-26
Refactor fewshot examples to use task generics instead of `**kwargs`

rmitsch:
## Description

- Refactor fewshot examples to use task generics instead of `**kwargs`.
- Remove Python 3.6 support.

\-

## Checklist

### DIFF
--- a/spacy_llm/tasks/rel/util.py
+++ b/spacy_llm/tasks/rel/util.py
@@ -1,9 +1,8 @@
-from typing import List
+import abc
+from typing import Generic, List
 
-from spacy.training import Example
-
-from ...compat import BaseModel, Self, validator
-from ...ty import FewshotExample
+from ...compat import BaseModel, validator
+from ...ty import FewshotExample, TaskContraT
 
 
 class RelationItem(BaseModel):
@@ -24,26 +23,9 @@ class EntityItem(BaseModel):
     label: str
 
 
-class RELExample(FewshotExample):
+class BaseRELExample(FewshotExample[TaskContraT], abc.ABC, Generic[TaskContraT]):
     text: str
     ents: List[EntityItem]
-    relations: List[RelationItem]
 
     class Config:
         arbitrary_types_allowed = True
-
-    @classmethod
-    def generate(cls, example: Example, **kwargs) -> Self:
-        entities = [
-            EntityItem(
-                start_char=ent.start_char,
-                end_char=ent.end_char,
-                label=ent.label_,
-            )
-            for ent in example.reference.ents
-        ]
-        return cls(
-            text=example.reference.text,
-            ents=entities,
-            relations=example.reference._.rel,
-        )
--- a/spacy_llm/tasks/rel/util.py
+++ b/spacy_llm/tasks/rel/util.py
@@ -1,8 +1,4 @@
-import abc
-from typing import Generic, List
-
 from ...compat import BaseModel, validator
-from ...ty import FewshotExample, TaskContraT
 
 
 class RelationItem(BaseModel):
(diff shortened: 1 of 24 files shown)


spacy_llm/tasks/rel/util.py
```diff
@@ -24,26 +23,9 @@ class EntityItem(BaseModel):
     label: str
 
 
-class RELExample(FewshotExample):
+class BaseRELExample(FewshotExample[TaskContraT], abc.ABC, Generic[TaskContraT]):
```
svlandeg, line 26: Why the renaming?

spacy_llm/tasks/span/examples.py
```diff
@@ -1,13 +1,17 @@
...
-from ...compat import BaseModel
-from ...ty import FewshotExample, Self
+from ...compat import BaseModel, Self
+from ...ty import FewshotExample, TaskContraT
 
 
-class SpanExample(FewshotExample, abc.ABC):
+class SpanExample(FewshotExample[TaskContraT], abc.ABC, Generic[TaskContraT]):
+    """Example for span tasks not using CoT.
+    Note: this should be SpanTaskContraT instead of TaskContraT, but this would entail a circular import.
```
svlandeg, line 12: Isn't there any other solution? A circular import typically means we should reorganize what is in which module...

spacy_llm/tasks/span/task.py
```diff
@@ -62,7 +63,7 @@ def __init__(
         self._description = description
 
         if self._prompt_examples:
-            self._prompt_examples = list(self._check_label_consistency(self))
+            self._prompt_examples = list(check_label_consistency(self))
```
svlandeg, line 66: I think it's fine to use the original code, refering to the method in `self`, as this was defined on L61. (imagine we'd change the definition on L61 in a future refactor, then we might miss that we're still using the "old" one here)

spacy_llm/ty.py
```diff
@@ -111,14 +102,24 @@ def parse_responses(
...
+TaskContraT = TypeVar("TaskContraT", bound=LLMTask, contravariant=True)
+
+
+class FewshotExample(GenericModel, abc.ABC, Generic[TaskContraT]):
+    @classmethod
+    @abc.abstractmethod
+    def generate(cls, example: Example, task: TaskContraT) -> Self:
+        """Create a fewshot example from a spaCy example.
+        example (Example): spaCy example.
+        task (_TaskContraT): Task for which to generate examples.
```
svlandeg, line 114:
```suggestion
        task (TaskContraT): Task for which to generate examples.
```

svlandeg (commented):
Mostly looks good!

As this removes 3.6 support, I'd target `develop` instead of `main` (though it might end up being the same release)

spacy_llm/tasks/span/examples.py
rmitsch: I agree, but in this case: 
- `SpanExample` should use `FewshotExample` parametrized with `SpanTaskContraT`: `FewshotExample[SpanTaskContraT]`.
- `SpanTaskContraT` requires `SpanTask` in its definition.
- `SpanTask` requires `SpanExample`.

This is a dependency loop, i. e. doing this in multiple files will (without local imports) always lead to circular import errors. The one alternative here would be to drop everything in `task.py` - also not nice :man_shrugging:

spacy_llm/tasks/rel/util.py
rmitsch: Outdated, removed it.

spacy_llm/tasks/span/task.py
rmitsch: Done in https://github.com/explosion/spacy-llm/pull/306/commits/19d3660926260ae1f159c782add4caeaf83476db.

PR #332, opened 2023-10-17
Support arbitrarily long docs

rmitsch:
## Description

Support arbitrarily long docs by implementing a map-reduce approach: split docs into shards, process each shard in a separate prompt, then fuse shards together using a consensus-finding algorithm.

Steps:
```
a. Do until until entire doc has been processed:  
       a. 1. Estimate n_tokens in rendered prompt  
       a. 2. If n_tokens within bounds: use complete doc as one shard  
       a. 3. Otherwise:  
           - Identify split-off point  
           - Split off first shard  
           - Repeat from a. 3. on with rest of the (yet unsharded) doc  
   b. For each shard: execute prompt, parse response  
   c. Fuse shards into one doc instance by applying a consenus mechanism.  
       Note: the consensus mechanism can be trivial - e. g. for NER: concat all .ents into new doc - or  
       not, e. g. for classification/regression tasks - what's a doc's class if there are three shards with  
       three different textcat results?  
```

todo

### Types of change

New feature.

## Checklist

### DIFF
--- a/spacy_llm/models/hf/base.py
+++ b/spacy_llm/models/hf/base.py
@@ -61,6 +61,13 @@ class HuggingFace(abc.ABC):
 
     @property
     @abc.abstractmethod
+    def context_length(self) -> int:
+        """Returns context length in number of tokens for this model.
+        RETURNS (int): Max. number of tokens in allowed in prompt for the current model.
+        """
+
+    @property
+    @abc.abstractmethod
     def hf_account(self) -> str:
         """Name of HF account for this model.
         RETURNS (str): Name of HF account.
--- a/spacy_llm/models/hf/falcon.py
+++ b/spacy_llm/models/hf/falcon.py
@@ -63,6 +63,10 @@ class Falcon(HuggingFace):
             default_cfg_run,
         )
 
+    @property
+    def context_length(self) -> int:
+        return 2048
+
 
 @registry.llm_models("spacy.Falcon.v1")
 def falcon_hf(
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [metadata]
-version = 0.6.1
+version = 0.6.2
 description = Integrating LLMs into structured NLP pipelines
 author = Explosion
 author_email = contact@explosion.ai
--- a/spacy_llm/models/hf/dolly.py
+++ b/spacy_llm/models/hf/dolly.py
@@ -46,6 +46,10 @@ class Dolly(HuggingFace):
             default_cfg_run,
         )
 
+    @property
+    def context_length(self) -> int:
+        return 2048
+
 
 @registry.llm_models("spacy.Dolly.v1")
 def dolly_hf(
--- a/spacy_llm/models/hf/llama2.py
+++ b/spacy_llm/models/hf/llama2.py
@@ -49,6 +49,10 @@ class Llama2(HuggingFace):
     def compile_default_configs() -> Tuple[Dict[str, Any], Dict[str, Any]]:
         return HuggingFace.compile_default_configs()
 
+    @property
+    def context_length(self) -> int:
+        return 4096
+
 
 @registry.llm_models("spacy.Llama2.v1")
 def llama2_hf(
--- a/spacy_llm/models/hf/mistral.py
+++ b/spacy_llm/models/hf/mistral.py
@@ -82,6 +82,10 @@ class Mistral(HuggingFace):
             default_cfg_run,
         )
 
+    @property
+    def context_length(self) -> int:
+        return 8000
+
 
 @registry.llm_models("spacy.Mistral.v1")
 def mistral_hf(
--- a/spacy_llm/models/hf/openllama.py
+++ b/spacy_llm/models/hf/openllama.py
@@ -76,6 +76,10 @@ class OpenLLaMA(HuggingFace):
             {**default_cfg_run, "max_new_tokens": 32},
         )
 
+    @property
+    def context_length(self) -> int:
+        return 2048
+
 
 @registry.llm_models("spacy.OpenLLaMA.v1")
 def openllama_hf(
--- a/spacy_llm/models/hf/stablelm.py
+++ b/spacy_llm/models/hf/stablelm.py
@@ -107,6 +107,10 @@ class StableLM(HuggingFace):
             },
         )
 
+    @property
+    def context_length(self) -> int:
+        return 4096
+
 
 @registry.llm_models("spacy.StableLM.v1")
 def stablelm_hf(
--- a/spacy_llm/models/rest/anthropic/model.py
+++ b/spacy_llm/models/rest/anthropic/model.py
@@ -54,7 +54,7 @@ class Anthropic(REST):
         headers = {
             **self._credentials,
             "model": self._name,
-            "anthropic_version": self._config.get("anthropic_version", "2023-06-01"),
+            "anthropic-version": self._config.get("anthropic-version", "2023-06-01"),
             "Content-Type": "application/json",
         }
 
--- a/spacy_llm/models/rest/anthropic/model.py
+++ b/spacy_llm/models/rest/anthropic/model.py
@@ -1,7 +1,7 @@
 import os
 import warnings
 from enum import Enum
-from typing import Any, Dict, Iterable, List, Sized, Tuple
+from typing import Any, Dict, Iterable, List, Sized
 
 import requests  # type: ignore[import]
 import srsly  # type: ignore[import]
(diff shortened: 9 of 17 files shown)


rmitsch:
Plumbing is done. Still TBD: mapping step for individual tasks.

spacy_llm/ty.py
```diff
@@ -85,20 +95,28 @@ def __call__(self, examples: Iterable[Example], **kwargs) -> Dict[str, Any]:
 
 @runtime_checkable
 class LLMTask(Protocol):
-    def generate_prompts(self, docs: Iterable[Doc]) -> Iterable[_PromptType]:
+    def generate_prompts(
+        self, docs: Iterable[Doc], context_length: Optional[int] = None
+    ) -> Iterable[Tuple[Iterable[_PromptType], Iterable[Doc]]]:
```
svlandeg, line 100: I think this new design makes sense, but I worry about the basic `LLMTask` becoming too complex for users to implement easily. Additionally, any currently existing user task will not work with `spacy-llm` anymore after this, which would be pretty unfortunate.

I wonder whether we could keep the best of both worlds: allow the more simple `LLMTask` as well as a new `ShardableLLMTask` (naming suggestions welcome). Would that be feasible?

rmitsch: I'm not a fan of a parallel task structure, but I see where you are coming from. I'll think about it.

rmitsch:
Update & remaining todos:
- [x] mapping/reducing for individual tasks.
- [ ] add tests for overlong documents
- [ ] currently we only transfer those custom attributes during splitting and reducing that are relevant to the task (e. g. `_.summary` for the summarization task) - this means that we lost custom attributes with a pipeline consisting of multiple LLM components writing out to custom attributes. We need to address that - as of now tasks are not aware of other tasks in the pipeline. Maybe we transfer all available custom attributes?

spacy_llm/ty.py
svlandeg: I guess the "shardable" one would be a sub class? It does offer additional functionality. It just feels wrong that we need to break the existing API to add those features, especially because not all user tasks might be implementing it.

rmitsch: So the concerns are about user tasks, right? Our built-in tasks would remain to be all shardable?
That makes sense, I misunderstood at first. Subclassing won't work here due to different function signatures. I think a different `Protocol` is the cleanest way to go.

I'll draft something.

rmitsch: Added a suggestion for this in https://github.com/explosion/spacy-llm/pull/332/commits/c104387059c18d0d79f8808ac50577e22b36998a. In a nutshell: we support shardable and non-shardable tasks now. Non-shardable ones look like they did before this PR, shardable ones have the nested prompt/response structure.

I chose to stick with the nested structure for models, as models are less likely to be user-contributed and it's handy from an implementation perspective if they can process the nested data structures. If we really want models to support both variants as well, we can do that, there's no strong reason not to (other than it doesn't add much of a benefit in my opinion).

spacy_llm/cache.py
```diff
@@ -68,11 +68,11 @@ def __init__(
 
         self._init_cache_dir()
 
-    def initialize(self, vocab: Vocab, task: LLMTask) -> None:
+    def initialize(self, vocab: Vocab, task: ShardingLLMTask) -> None:
```
svlandeg, line 71: The cache is still generic enough to also work with the normal task types, no? i.e. `Union[NonshardingLLMTask, ShardingLLMTask]`?

spacy_llm/ty.py
```diff
@@ -84,7 +93,34 @@ def __call__(self, examples: Iterable[Example], **kwargs) -> Dict[str, Any]:
 
 
 @runtime_checkable
-class LLMTask(Protocol):
```
svlandeg, line 87: Can we keep the definition of `LLMTask` here as the union between `ShardingLLMTask` and `NonshardingLLMTask`? That would avoid some edits in the test suite - the tasks would still be instances of `LLMTask`, as before.

spacy_llm/ty.py
svlandeg: So if a user currently has an `LLMTask` implemented, and they'd update `spacy-llm` with this PR merged as is, would the built-in models still support that old task variant?

spacy_llm/models/hf/base.py
```diff
@@ -39,10 +39,10 @@ def __init__(
         self._model = self.init_model()
 
     @abc.abstractmethod
-    def __call__(self, prompts: Iterable[Any]) -> Iterable[Any]:
+    def __call__(self, prompts: Iterable[Iterable[Any]]) -> Iterable[Iterable[Any]]:
```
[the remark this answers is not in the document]
svlandeg: Hm right, if we do it in an outer loop then you basically prevent the model to create batches that span across docs, right? Your batch would be maximum one doc, ie one set of shards?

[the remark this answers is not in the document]
rmitsch: Exactly.

svlandeg (commented):
Haven't been able to review everything yet, but left some comments and questions already to help me understand some of the design decisions.

spacy_llm/models/hf/base.py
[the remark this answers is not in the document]
svlandeg: right.

spacy_llm/pipeline/llm.py
```diff
@@ -34,6 +36,7 @@
 
 DEFAULT_SAVE_IO = False
 DEFAULT_VALIDATE_TYPES = True
+_LLMTask = Union[NonshardingLLMTask, ShardingLLMTask]
```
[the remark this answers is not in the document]
rmitsch: Moved in https://github.com/explosion/spacy-llm/pull/332/commits/f40bc88217ebe9c02e7bef9d473670d940fa338b.

spacy_llm/models/hf/base.py
```diff
@@ -17,16 +17,18 @@ def __init__(
...
+        context_length: int,
     ):
         """Initializes HF model instance.
         query (Callable[[Any, Iterable[Any]], Iterable[Any]): Callable executing LLM prompts when
             supplied with the `integration` object.
         name (str): Name of HF model to load (without account name).
         config_init (Optional[Dict[str, Any]]): HF config for initializing the model.
         config_run (Optional[Dict[str, Any]]): HF config for running the model.
-        inference_config (Dict[Any, Any]): HF config for model run.
+        context_length (int): Context length for this model. Necessary for sharding.
```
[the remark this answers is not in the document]
rmitsch: Updated in https://github.com/explosion/spacy-llm/pull/332/commits/f40bc88217ebe9c02e7bef9d473670d940fa338b.

spacy_llm/ty.py
rmitsch: Yes.

svlandeg (commented):
This mostly looks good to me, and in particular the "reduce" behaviour of the tasks all seem sensible. Will certainly be worth the effort to have this!

spacy_llm/pipeline/llm.py
```diff
@@ -68,7 +70,7 @@ def make_llm(
     nlp (Language): Pipeline.
     name (str): The component instance name, used to add entries to the
         losses during training.
-    task (Optional[LLMTask]): An LLMTask can generate prompts for given docs, and can parse the LLM's responses into
+    task (Optional[_LLMTask]): An _LLMTask can generate prompts for given docs, and can parse the LLM's responses into
```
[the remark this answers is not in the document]
rmitsch: Fixed in https://github.com/explosion/spacy-llm/pull/332/commits/63367fab55cc69e5b1872b1ccd578740a7443744.

spacy_llm/models/hf/falcon.py
```diff
@@ -17,9 +17,15 @@ def __init__(
         name: MODEL_NAMES,
         config_init: Optional[Dict[str, Any]],
         config_run: Optional[Dict[str, Any]],
+        context_length: int,
```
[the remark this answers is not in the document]
rmitsch: Done in https://github.com/explosion/spacy-llm/pull/332/commits/63367fab55cc69e5b1872b1ccd578740a7443744.

spacy_llm/models/hf/base.py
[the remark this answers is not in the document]
rmitsch: Fixed in https://github.com/explosion/spacy-llm/pull/332/commits/63367fab55cc69e5b1872b1ccd578740a7443744.

rmitsch:
> I guess one optimization could be doing something like a "consistency" check, running the aggregated response once more through an LLM to optimize for fluency (thinking about Summarization & Translation tasks for instance). But I think that can be a future refinement.

Agreed on both points. I can totally see how that would improve the result, but wouldn't include it in this PR.

spacy_llm/pipeline/llm.py
```diff
@@ -206,24 +210,60 @@ def _process_docs(self, docs: List[Doc]) -> List[Doc]:
...
+            context_length: Optional[int] = None
+            if isinstance(self._model, ModelWithContextLength):
+                context_length = self._model.context_length
+            if has_shards and context_length is None:
+                warnings.warn(
+                    "Task supports sharding, but model does not provide context length. Data won't be sharded, prompt "
+                    "might exceed the model's context length. Set context length in your config. If you think spacy-llm"
+                    " should provide the context length for this models automatically, report this to "
+                    "https://github.com/explosion/spacy-llm/issues."
+                )
```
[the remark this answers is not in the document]
svlandeg: Let's keep it at `warn`.

spacy_llm/cache.py
```diff
@@ -72,7 +72,7 @@ def initialize(self, vocab: Vocab, task: LLMTask) -> None:
         """
         Initialize cache with data not available at construction time.
         vocab (Vocab): Vocab object.
-        task (LLMTask): Task.
+        task (ShardingLLMTask): Task.
```
svlandeg, line 75:
```suggestion
        task (LLMTask): Task.
```

svlandeg (approved):
🥳

PR #359, opened 2023-11-09
Improve HF device handling

rmitsch:
## Description

Improve HF device handling:
- Consider device/device_map conflicts. 
- Always move inputs to model device for models using AutoModelForCausalLM.

Context/motivation: https://github.com/explosion/spacy-llm/discussions/324#discussioncomment-7338342, https://github.com/explosion/spacy-llm/discussions/324. This PR should help with both.

\-

## Checklist

### DIFF
--- a/spacy_llm/models/hf/base.py
+++ b/spacy_llm/models/hf/base.py
@@ -53,7 +53,11 @@ class HuggingFace(abc.ABC):
                     self._config_init.pop("device")
 
         # Fetch proper torch.dtype, if specified.
-        if has_torch and self._config_init.get("torch_dtype", "") not in ("", "auto"):
+        if (
+            has_torch
+            and "torch_dtype" in self._config_init
+            and self._config_init["torch_dtype"] != "auto"
+        ):
             try:
                 self._config_init["torch_dtype"] = getattr(
                     torch, self._config_init["torch_dtype"]
--- a/spacy_llm/models/hf/base.py
+++ b/spacy_llm/models/hf/base.py
@@ -124,7 +124,7 @@ class HuggingFace(abc.ABC):
         default_cfg_run: Dict[str, Any] = {}
 
         if has_torch:
-            default_cfg_init["torch_dtype"] = torch.bfloat16
+            default_cfg_init["torch_dtype"] = "bfloat16"
             if has_torch_cuda_gpu:
                 # this ensures it fails explicitely when GPU is not enabled or sufficient
                 default_cfg_init["device"] = "cuda:0"
--- a/spacy_llm/models/hf/base.py
+++ b/spacy_llm/models/hf/base.py
@@ -27,11 +27,30 @@ class HuggingFace(abc.ABC):
         inference_config (Dict[Any, Any]): HF config for model run.
         """
         self._name = name if self.hf_account in name else f"{self.hf_account}/{name}"
-        self._config_init, self._config_run = self.compile_default_configs()
+        default_cfg_init, default_cfg_run = self.compile_default_configs()
+        self._config_init, self._config_run = default_cfg_init, default_cfg_run
+
         if config_init:
             self._config_init = {**self._config_init, **config_init}
         if config_run:
             self._config_run = {**self._config_run, **config_run}
+
+        # `device` and `device_map` are conflicting arguments - ensure they aren't both set.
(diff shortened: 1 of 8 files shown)


adrianeboyd:
I do think it's a good idea to test with `accelerate`, at least in the future. It looks like that torch bug will be fixed in torch v2.1.1.

rmitsch:
> I do think it's a good idea to test with `accelerate`, at least in the future. It looks like that torch bug will be fixed in torch v2.1.1.

I'll do that in a follow-up PR as I want this to be included in the upcoming v0.6.3 release (release today or tomorrow).

rmitsch:
> I do think it's a good idea to test with `accelerate`, at least in the future. It looks like that torch bug will be fixed in torch v2.1.1.

I'll do that in a follow-up PR as I want this to be included in the upcoming v0.6.3 release (release today or tomorrow).

rmitsch:
> I do think it's a good idea to test with `accelerate`, at least in the future. It looks like that torch bug will be fixed in torch v2.1.1.

I'll do that in a follow-up PR as I want this to be included in the upcoming v0.6.3 release (release today or tomorrow).

spacy_llm/models/hf/base.py
```diff
@@ -27,12 +27,47 @@ def __init__(
...
+            if "device_map" in default_cfg_init and "device" in config_init:
+                self._config_init.pop("device_map")
+            # Case 3: both explicitly set by user.
+            if "device" in config_init and "device_map" in config_init:
+                warnings.warn(
+                    "`device` and `device_map` are conflicting arguments - don't set both. Dropping argument "
+                    "`device`."
+                )
+                if "device" in self._config_init:
+                    self._config_init.pop("device")
```
svlandeg, line 53: I would swap these cases and make the current Case 3 the first `if` check. The reason is that if the user has both defined in their `config_init`, the behaviour of which is removed shouldn't depend on previous checks against the `default_cfg_init`, and we'd probably want to warn explicitely about the conflict so they understand that one might be removed. Also by having it as first condition, you can remove the extra check `if "device" in self._config_init`. The other two can then be `elif`'s.

spacy_llm/tests/models/test_hf.py
```diff
@@ -0,0 +1,80 @@
+from typing import Tuple
+
+import pytest
+import spacy
+from thinc.compat import has_torch_cuda_gpu
+
+from spacy_llm.compat import has_accelerate
+
+from ...compat import torch
```
svlandeg, line 9:
```suggestion
from spacy_llm.compat import has_accelerate, torch
```

svlandeg:
There's no documentation update necessary after this?

rmitsch:
> There's no documentation update necessary after this?

I wouldn't think so - this is part bugfix and part the ability to process `torch_dtype` in the config, which should have been there anyway, as we claim all HF args can be set and are passed on to the HF model. I'd say describing that in the release notes should be sufficient, but we can make this explicit in the docs as well.

spacy_llm/models/hf/base.py
rmitsch: Done in https://github.com/explosion/spacy-llm/pull/359/commits/17a6402c2f2a230a046a95a59637a4e0fd30e363.

spacy_llm/models/hf/base.py
```diff
@@ -27,12 +27,46 @@ def __init__(
...
         """
         self._name = name if self.hf_account in name else f"{self.hf_account}/{name}"
-        self._config_init, self._config_run = self.compile_default_configs()
+        default_cfg_init, default_cfg_run = self.compile_default_configs()
+        self._config_init, self._config_run = default_cfg_init, default_cfg_run
+
         if config_init:
             self._config_init = {**self._config_init, **config_init}
         if config_run:
             self._config_run = {**self._config_run, **config_run}
```
svlandeg, line 36: Still wrapping my head around how this actually works. Shouldn't we only add keys from the `default_X` dictionaries that aren't already in either `config_init` or `config_run`, instead of just plainly merging these?

rmitsch: The entries in the second dict overwrite the entries in the first dict - so e. g. `{**{"a": 1, "b": 2}, **{"a": 3, "c": 4}} == {'a': 3, 'b': 2, 'd': 4}`. Meaning that if we have something in our default config, user-specified entries override those from the default config. This should be identical to only adding values from the default config that aren't in the user-provided config.

PR #396, opened 2023-12-05
Add `TranslationTask`

rmitsch:
## Description

Add `TranslationTask`.

https://github.com/explosion/spaCy/pull/13183

### Types of change

New feature.

## Checklist


spacy_llm/tasks/translation/registry.py
```diff
@@ -0,0 +1,59 @@
...
+def make_translation_task(
+    target_lang: str,
+    source_lang: Optional[str] = None,
+    template: str = DEFAULT_TRANSLATION_TEMPLATE_V1,
+    parse_responses: Optional[TaskResponseParser[TranslationTask]] = None,
+    prompt_example_type: Optional[Type[FewshotExample]] = None,
+    examples: ExamplesConfigType = None,
+    shard_mapper: Optional[ShardMapper] = None,
+    shard_reducer: Optional[ShardReducer] = None,
+    field: str = "translation",
```
svlandeg, line 27: I wonder whether we should make the default depending on the target language, basically making it easier for users to have multiple translation components for different target languages. Though I suppose in that case they can just set the field themselves.

rmitsch: You mean naming the field `"translation_{self._target_lang}"` by default?

svlandeg: Yes, something like that. You'd have to adjust the default value in the function body though. So it's probably not worth the hassle?

rmitsch: Probably not. Alternative suggestion: a dict with language names. That would allow easier/automatic merging for multiple translation components with different languages. I. e. instead of `self._translation = ...` we'd have 
```python
self._translation = {
    self._target_lang: ...
}
```
What do you think? Otherwise I'd stick with the current solution.

svlandeg: I think that'll make things too complex for users wanting just a single translation. Let's stick with the current implementation :-)

PR #434, opened 2024-01-29
Fix/index error when unhighlighting

magdaaniol:
## Description

The logic in `unhighlight_ents_in_doc` does not account for the fact that there might exist unhighlighted entities in doc, which will lead to IndexError if that's the case.

### Types of change

Bug fix

## Checklist


rmitsch (approved):
Thanks for the fix! :tada:

PR #501, opened 2026-03-24
Add MiniMax as LLM provider backend

octo-patch:
## Summary

Adds [MiniMax](https://platform.minimaxi.com/) as a first-class REST LLM provider backend for spacy-llm, alongside OpenAI, Anthropic, Cohere, and PaLM.

MiniMax offers an OpenAI-compatible chat completions API with models like **MiniMax-M2.5** (204K context) and **MiniMax-M2.7** (1M context), making it a compelling alternative for structured NLP pipelines.

## Changes

- **New provider module** `spacy_llm/models/rest/minimax/` with model, registry, and init files
- **Registry entry** `spacy.MiniMax.v1` for seamless spaCy config integration
- **Think-tag stripping** for reasoning model output
- **Context length definitions** for M2.5, M2.5-highspeed, M2.7, M2.7-highspeed
- **Test suite** with unit tests (context lengths, endpoints) and integration tests (API response validation, error handling)
- **Usage example** config for text classification with MiniMax
- **README** updated to list MiniMax as a supported provider

### Python
```python
import spacy

nlp = spacy.blank("en")
nlp.add_pipe(
    "llm",
    config={
        "model": {
            "@llm_models": "spacy.MiniMax.v1",
            "name": "MiniMax-M2.5",
            "config": {"temperature": 0.0},
        },
        "task": {
            "@llm_tasks": "spacy.TextCat.v2",
            "labels": ["POSITIVE", "NEGATIVE"],
        },
    },
)
doc = nlp("This is a great product!")
print(doc.cats)
```

### Config file
```ini
[components.llm.model]
@llm_models = "spacy.MiniMax.v1"
name = "MiniMax-M2.5"
config = {"temperature": 0.0}
```

## Supported Models

| Model | Context Length |
|-------|---------------|
| MiniMax-M2.7 | 1,048,576 |
| MiniMax-M2.7-highspeed | 1,048,576 |
| MiniMax-M2.5 | 204,800 |
| MiniMax-M2.5-highspeed | 204,800 |

## Test Plan

- [x] Unit tests pass: context length definitions, endpoint constants
- [x] Integration tests pass: API response validation with M2.5 and M2.5-highspeed
- [x] Error handling tests pass: unsupported model, invalid config
- [x] Registry integration verified: `spacy.MiniMax.v1` resolves correctly
- [x] Existing tests unaffected

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ This package integrates Large Language Models (LLMs) into [spaCy](https://spacy.
   - **[Anthropic](https://docs.anthropic.com/claude/reference/)**
   - **[Google PaLM](https://ai.google/discover/palm2/)**
   - **[Microsoft Azure AI](https://azure.microsoft.com/en-us/solutions/ai)**
+  - **[MiniMax](https://platform.minimaxi.com/)**
 - Supports open-source LLMs hosted on Hugging Face 🤗:
   - **[Falcon](https://huggingface.co/tiiuae)**
   - **[Dolly](https://huggingface.co/databricks)**
--- a/spacy_llm/models/__init__.py
+++ b/spacy_llm/models/__init__.py
@@ -1,10 +1,11 @@
 from .hf import dolly_hf, openllama_hf, stablelm_hf
 from .langchain import query_langchain
-from .rest import anthropic, cohere, noop, openai, palm
+from .rest import anthropic, cohere, minimax, noop, openai, palm
 
 __all__ = [
     "anthropic",
     "cohere",
+    "minimax",
     "openai",
     "dolly_hf",
     "noop",
--- a/spacy_llm/models/rest/__init__.py
+++ b/spacy_llm/models/rest/__init__.py
@@ -1,10 +1,11 @@
-from . import anthropic, azure, base, cohere, noop, openai
+from . import anthropic, azure, base, cohere, minimax, noop, openai
 
 __all__ = [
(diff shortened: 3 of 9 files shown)


JiwaniZakir (commented):
In `model.py`, the error-handling path inside the `__call__` loop has a type mismatch bug: when `"error" in responses`, `return responses["error"]` returns a flat `List[str]` (e.g., `["<json error>", "<json error>"]`), but the declared return type is `Iterable[Iterable[str]]`. Callers expecting a nested structure will silently iterate over individual characters of the error string rather than getting a proper response list. This should be `return [responses["error"]]` or appended to `all_api_responses` before returning, consistent with the non-error path.

Additionally, `_verify_auth` in `model.py` verifies credentials by making a live API call with `self([["test"]])`, which consumes billable tokens on every cold start or config validation. The other providers in this repo (e.g., OpenAI's implementation) typically use a lightweight `/v1/models` endpoint for this. Since MiniMax reportedly lacks that endpoint, it may be worth at minimum documenting this side-effect in a comment, or skipping the verification entirely and relying on the first real call to surface auth errors.

Finally, the `_request` closure is redefined on every iteration of the outer `for prompts_for_doc in prompts` loop — moving it outside the loop (accepting `prompts_for_doc` as a parameter) would be cleaner and avoids the closure capturing a loop variable.

